PR #11, opened 2021-04-08
feature/sdk support serve provider

roneli:


proto/internal/generate.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
yevgenypats, line 1: Just curious - Is it possible to use the go generate ?

serve/serve.go
```diff
@@ -0,0 +1,98 @@
...
+			// be filtered.
+			Level:      hclog.Trace,
+			JSONFormat: true,
+		})
+		log.SetOutput(logger.StandardWriter(&hclog.StandardLoggerOptions{InferLevels: true}))
+	}
+	plugin.Serve(&plugin.ServeConfig{
+		HandshakeConfig: Handshake,
+		VersionedPlugins: map[int]plugin.PluginSet{
+			1: {
```
yevgenypats, line 89: is it still the same version? shouldn't we increase it to 2 ?

yevgenypats (changes requested):

serve/serve.go
roneli: still same version

yevgenypats (approved):

Issue #9, opened 2021-04-04
disallow column longer then 63 characters


yevgenypats:

roneli:
and tables*

roneli:
- Verify columns names
- Verify table names

Issue #7, opened 2021-04-04
Add an error when a user sets a field that doesn't exist


yevgenypats:
The SDK should panic or at least return an error when `resource.Set('column', value)` to a `column that is not defined`

PR #20, opened 2021-05-03
Added resource set validation

zagronitay:
Added resource set validation https://github.com/cloudquery/cq-provider-sdk/issues/7

Added TableValidation to enforce column/table name longer than 63 char https://github.com/cloudquery/cq-provider-sdk/issues/9


provider/migrator.go
```diff
@@ -109,3 +124,33 @@ func (m Migrator) buildColumns(ctb *sqlbuilder.CreateTableBuilder, cc []schema.C
...
 }
+
+func (m Migrator) ValidateTable(t *schema.Table) error {
+	for _, validator := range m.validators {
+		return validator.Validate(t)
+	}
+	return nil
+}
+
+type TableValidator interface {
```
roneli, line 135: move validators to their own file i.e validators.go

provider/migrator.go
```diff
@@ -109,3 +124,33 @@ func (m Migrator) buildColumns(ctb *sqlbuilder.CreateTableBuilder, cc []schema.C
...
+	Validate(t *schema.Table) error
+}
+
+type LengthTableValidator struct {
+	log hclog.Logger
+}
+
+func (tv LengthTableValidator) Validate(t *schema.Table) error {
+	if len(t.Name) > maxTableName {
```
roneli, line 144: what about the table's relations? and their relations?

provider/provider.go
```diff
@@ -60,6 +60,14 @@ func (p *Provider) Init(_ string, dsn string, _ bool) error {
 	// Create tables
 	m := NewMigrator(p.db, p.Logger)
 	for _, t := range p.ResourceMap {
+
+		// validate table
+		validationErr := m.ValidateTable(t)
```
roneli, line 65: maybe its unnecessary to have the table validator part of the migrator, it does have a different purpose. We can have a defaultTableValidator and use that for now, or even just a function, and move that into a struct in the future.

provider/schema/resource.go
```diff
@@ -32,6 +34,10 @@ func (r Resource) Get(key string) interface{} {
 }
 
 func (r Resource) Set(key string, value interface{}) {
+	columnExists := funk.ContainsString(r.table.ColumnNames(), key)
+	if !columnExists {
+		panic(fmt.Sprintf("column %s does not exits", key))
```
roneli, line 39: Exists*, and maybe it is better to return an error here instead (I know it will basically means we have to update all providers to check / return error), it is better than a panic imo. Providers can update to this new syntax later.

roneli (changes requested):

provider/schema/resolvers.go
```diff
@@ -9,13 +9,19 @@ import (
 
 func PathResolver(path string) ColumnResolver {
 	return func(_ context.Context, meta ClientMeta, r *Resource, c Column) error {
-		r.Set(c.Name, funk.GetAllowZero(r.Item, path))
+		err := r.Set(c.Name, funk.GetAllowZero(r.Item, path))
```
roneli, line 12: why not just return?

provider/schema/resolvers.go
```diff
@@ -9,13 +9,19 @@ import (
...
+		if err != nil {
+			return err
+		}
 		return nil
 	}
 }
 
 func ParentIdResolver(_ context.Context, _ ClientMeta, r *Resource, c Column) error {
-	r.Set(c.Name, r.Parent.Id())
+	err := r.Set(c.Name, r.Parent.Id())
```
roneli, line 21: same as above why not return?

roneli (changes requested):

roneli (approved):

PR #31, opened 2021-05-31
Support Network columns Closes #22

### LINKED ISSUE #22: Support Network column types
Support the following column types:

- [x] cidr
- [x] mac
- [x] inet

roneli:


provider/schema/column.go
```diff
@@ -95,6 +114,18 @@ func ValueTypeFromString(s string) ValueType {
 		return TypeTimestamp
 	case "uuidarray", "TypeUUIDArray":
 		return TypeUUIDArray
+	case "typeinet", "TypeInet":
```
michelvocks, line 117: Nit: I guess it is not necessary to check other cases than lower case since you cast it to lower case at the top.

michelvocks (approved):
LGTM. Should be ready for merge after the liniting issue has been fixed.

provider/schema/column.go
roneli: fixed shouldn't have the prefix type

PR #34, opened 2021-06-10
feat: handle []interface{}, interface{} for TypeJSON; add TypeInterval

### LINKED ISSUE #36: JSON column validation rejects []interface{}
`TypeJSON` columns currently only accept `map`, `[]byte`, `string,` and `*string` values: https://github.com/cloudquery/cq-provider-sdk/blob/main/provider/schema/column.go#L139
```go
	// Maps are jsons
	if reflect2.TypeOf(v).Kind() == reflect.Map {
		return c.Type == TypeJSON
	}

	switch val := v.(type) {
	...
	case []byte:
		...
		return c.Type == TypeByteArray || c.Type == TypeJSON
	...
	case string:
		...
		if c.Type == TypeJSON {
			return true
		}
		...
	case *string:
		if c.Type == TypeJSON {
			return true
		}
```
... whereas postgres `json` & `jsonb` columns support JSON arrays as well as primitives and pgx accepts virtually anything it can marshal: https://github.com/jackc/pgtype/blob/master/json.go#L52-L57

at the very least it would be useful to be store JSON-serializable slices in `TypeJSON` columns

### LINKED ISSUE #37: add support for postgres interval columns
I propose adding a column type `TypeInterval` for postgres `interval` given pgx's existing support: https://github.com/jackc/pgtype/blob/master/interval.go

obataku:
closes #36
closes #37 

adds support for storing non-object data in JSON columns (e.g. arrays) as well as pg `interval` columns (pgx supports it anyway)

### DIFF
--- a/provider/schema/column.go
+++ b/provider/schema/column.go
@@ -28,6 +28,7 @@ const (
 	TypeStringArray
 	TypeIntArray
 	TypeTimestamp
+	TypeInterval
 	TypeJSON
 	TypeUUIDArray
 )
@@ -56,6 +57,8 @@ func (v ValueType) String() string {
 		return "TypeStringArray"
 	case TypeTimestamp:
 		return "TypeTimestamp"
+	case TypeInterval:
+		return "TypeInterval"
 	case TypeByteArray:
 		return "TypeByteArray"
 	case TypeUUIDArray:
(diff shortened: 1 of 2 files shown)


roneli:
Sorry for the late response. Seem like two great types to support, need to do one check on my end regarding adding new types. 

One last thing is required since adding a new type TypeInterval also requires to update the plugin protocol (see [cqproto ](https://github.com/cloudquery/cq-provider-sdk/blob/38188fd6dcadf0ce53a4335d73e442ac7f0fb8fd/cqproto/internal/plugin.proto#L85)) 

So we need both enums to support the same order of values.

roneli (changes requested):
See comment above

roneli:
#36 Supported in #50

PR #41, opened 2021-06-24
SDK Improvements

roneli:
- add support to disable delete
- move table creation outside of SDK. caller should create tables
- changed id to cq_id (more readable)
-  add meta field
- add support for user defined primary keys


cqproto/grpc.go
```diff
@@ -102,13 +109,20 @@ func (g *GRPCServer) GetProviderConfig(ctx context.Context, _ *internal.GetProvi
 }
 
 func (g *GRPCServer) ConfigureProvider(ctx context.Context, request *internal.ConfigureProvider_Request) (*internal.ConfigureProvider_Response, error) {
+
+	var eFields = make(map[string]interface{})
+	if err := json.Unmarshal(request.GetExtraFields(), &eFields); err != nil {
```
michelvocks, line 114: Will `GetExtraFields` ever return `nil`? If so, I think this will panic here.

cqproto/internal/plugin.proto
```diff
@@ -26,6 +25,10 @@ message ConfigureProvider {
     ConnectionDetails connection = 2;
     // Holds information such as credentials, regions, accounts, etc'
     bytes config = 3;
+    // Disables auto delete of table data before fetch of resources
+    bool disableDelete = 4;
+    // Allows to inject inject & override fields into resource tables, use this carefully to override fields
```
michelvocks, line 30:
```suggestion
    // Allows to inject & override fields into resource tables, use this carefully to override fields
```

provider/schema/meta.go
```diff
@@ -0,0 +1,14 @@
+package schema
+
+var (
+	cqIdColumn = Column{
+		Name:        "cq_id",
+		Type:        TypeUUID,
+		Description: "Unique CloudQuery Id given to the resource",
```
michelvocks, line 7:
```suggestion
		Description: "Unique CloudQuery Id added to every resource",
```

provider/schema/resource.go
```diff
@@ -16,47 +18,97 @@ type Resource struct {
...
+
+func (r *Resource) GenerateCQId() error {
+	if len(r.table.Options.PrimaryKeys) == 0 {
+		return nil
+	}
+	objs := make([]interface{}, len(r.table.PrimaryKeys()))
+	for i, pk := range r.table.PrimaryKeys() {
+		value := r.Get(pk)
+		if value == nil {
+			return fmt.Errorf("failed to generate cq_id for %s, pk field missing %s", r.table.Name, pk)
```
michelvocks, line 81: Is it necessary to fail and exit here or do you think it may makes more sense to continue without generating the cq_id and printing out a warning message?

michelvocks (approved):
LGTM from a high level view. I will do a more comprehensive test tomorrow.

cqproto/grpc.go
roneli: https://play.golang.org/p/dyqF3cZMWOd
won't panic but will return an error which is not the flow we want.
fixed to check for nil

provider/schema/resource.go
roneli: I am not sure, basically this should almost never happen since a PK can't be null, if it won't fail here it will fail in the table insert

PR #48, opened 2021-07-13
Add copy from insert

roneli:
- add support for copy from (deletes and cascades on main table)
- extra fields use msgpack marshal instead of json.


provider/schema/database.go
```diff
@@ -68,6 +77,51 @@ func (p PgDatabase) Insert(ctx context.Context, t *Table, resources []*Resource)
...
+		if err != nil {
+			return err
+		}
+		if copied != int64(len(resources)) {
+			return fmt.Errorf("not all resources copied %d != %d to %s", copied, len(resources), resources.TableName())
+		}
+
+		return nil
+	})
+	if err != nil {
```
michelvocks, line 119: this can be simplified by just returning `err`

provider/schema/database.go
```diff
@@ -159,3 +213,45 @@ func TruncateTableConstraint(name string) string {
...
+		if err := c.ValidateType(v); err != nil {
+			return nil, err
+		}
+		values = append(values, v)
+		if c.Type == TypeJSON {
+			if reflect2.TypeOf(v).Kind() == reflect.Map {
+				values = append(values, v)
+				continue
+			}
+			switch data := v.(type) {
```
michelvocks, line 230: The downside of this is that we have to keep track of all the value types that could be used inside a database, am I right with that assumption? I don't see another way to solve this so I guess it is fine but it might be worth keep this in our heads for the future.

michelvocks (approved):
LGTM. One small suggestion and one question.

provider/schema/database.go
roneli: Yes you are right, but its more related to pgx and how it encodes data from what I saw its mostly only json that is required to be a map[ type

PR #51, opened 2021-07-15
Fix nil json

fdistorted:


provider/schema/database.go
```diff
@@ -219,6 +219,9 @@ func getResourceValues(r *Resource) ([]interface{}, error) {
 			return nil, err
 		}
 		if c.Type == TypeJSON {
+			if v == nil {
+				continue
```
roneli, line 223: you need to append the value other wise column count != value count

provider/schema/database_json_test.go
```diff
@@ -7,36 +7,30 @@ import (
 )
 
 var (
```
roneli, line 9: why not put the tests inside database_test.go?

roneli (changes requested):
Minor change request

roneli (approved):

PR #62, opened 2021-08-03
added more default resolvers

fdistorted:


provider/schema/resolvers.go
```diff
@@ -38,3 +42,139 @@ func ParentPathResolver(path string) ColumnResolver {
...
+
+// DateResolver resolves the different date formats (ISODate - 2011-10-05T14:48:00.000Z is default) into *time.Time
+//
+// Examples:
+// DateResolver("Date", "") - resolves using RFC.RFC3339 as default
+//
+// DateResolver("InnerStruct.Field", time.RFC822)  - resolves using time.RFC822
+//
+// DateResolver("InnerStruct.Field", "2011-10-05")  - resolves using manually set format
+func DateResolver(path string, rfc string) ColumnResolver {
```
roneli, line 54: I would add support for multiple rfcs with ...string so all rfcs are tried by that order

roneli, line 54: Would also create a  DateUTCResolver that converts to date and then into UTC

provider/schema/resolvers_test.go
```diff
@@ -68,3 +71,170 @@ func TestInterfaceSlice(t *testing.T) {
...
+	},
+}
+
+type testNetStruct struct {
+	IP  string
+	MAC string
+	Net string
+}
+
+func TestNetResolvers(t *testing.T) {
```
roneli, line 142: - [ ] Test also error cases
- [ ] Test ipv6, ip passed with subnet, etc'

roneli (changes requested):
Looks good a few changes needed

roneli (approved):
LGTM

PR #64, opened 2021-08-05
json resolver adjusted

fdistorted:
now sdk tries to parse everything into json
fixes cloudquery/cloudquery-issues#82


provider/schema/column.go
```diff
@@ -267,6 +267,6 @@ func (c Column) checkType(v interface{}) bool {
 			return true
 		}
 	}
-
-	return false
+	// everything can be a json
+	return c.Type == TypeJSON
```
roneli, line 271: I disagree common types aren't allowed to be jsons, only structs, maps, arrays etc'

provider/schema/database.go
```diff
@@ -254,6 +254,17 @@ func getResourceValues(r *Resource) ([]interface{}, error) {
 					return nil, err
 				}
 				values = append(values, newV)
+			default:
```
roneli, line 257: I think passing a "string" as json to the copy from will result in a warning and trigger normal bulk insert, we don't want that, hence my top comment

roneli (changes requested):
Some questions about the resolver

roneli (approved):
LGTM

PR #66, opened 2021-08-06
Generate cq_id

### LINKED ISSUE #63: Generate random cq_id if some primary keys are null
Currently if any primary key is null we return an error when doing GenerateCqID, instead of returning an error log a warning and and generate a random ID instead, this should only be allowed for tables that have a parent, main tables must always return an error.

fdistorted:
generates random cq_id  for child resources if there are `nil` primary keys 
fixes #63


roneli:
closed in #68

PR #87, opened 2021-09-06
support ["*"] as all resources

roneli:
closes #87 

makes enable_partial_fetch enabled by default

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -72,7 +72,10 @@ func (p *Provider) GetProviderConfig(_ context.Context, _ *cqproto.GetProviderCo
 	data := fmt.Sprintf(`
 		provider "%s" {
 			%s
+			// list of resources to fetch
 			resources = %s
+			// enables partial fetching, allowing for any failures to not stop full resource pull
+			enable_partial_fetch = true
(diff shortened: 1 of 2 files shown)


provider/provider.go
```diff
@@ -176,6 +182,20 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
...
 }
 
+func (p *Provider) interpolateAllResources(requestedResources []string) []string {
+	if len(requestedResources) != 1 {
+		return requestedResources
+	}
+	if requestedResources[0] != "*" {
+		return requestedResources
+	}
+	allResources := make([]string, 0)
```
disq, line 192:
```suggestion
	allResources := make([]string, 0, len(p.ResourceMap))
```
(could also use a counter variable and do `allResources[i] = k; i++` and get rid of the `append`...)

fdistorted (approved):
Looks good.

provider/provider.go
roneli: added the cap, I like the append feels more cleaner than adding a counter.

disq (approved):

PR #95, opened 2021-09-19
cleanup stale data after fetch

roneli:
Fixes https://github.com/cloudquery/cloudquery/issues/199

- Fixed Edge case where a resource failed resolving because of a column/post/sdk resolving error. This would insert a bad resource and cause errors.
- Added IgnoreFunc to column structs

### DIFF
--- a/provider/schema/execution.go
+++ b/provider/schema/execution.go
@@ -252,7 +252,7 @@ func (e *ExecutionData) copyDataIntoDB(ctx context.Context, resources Resources,
 	}
 	e.Logger.Warn("failed copy-from to db", "error", err, "table", e.Table.Name)
 
-	// fallback insert, copy from sometimes does problems so we fall back with insert
+	// fallback insert, copy from sometimes does problems, so we fall back with bulk insert
 	err = e.Db.Insert(ctx, e.Table, resources)
 	if err == nil {
 		return resources, nil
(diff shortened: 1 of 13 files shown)


yevgenypats (approved):

provider/schema/execution.go
```diff
@@ -62,6 +67,7 @@ func NewExecutionData(db Database, logger hclog.Logger, table *Table, disableDel
 		extraFields:               extraFields,
 		PartialFetchFailureResult: []PartialFetchFailedResource{},
 		partialFetch:              partialFetch,
+		executionStart:            time.Now().Add(-1 * time.Minute),
```
yevgenypats, line 70: What does it mean ? maybe const ?

yevgenypats: Like why execution started 1 min ago ?

provider/schema/execution.go
```diff
@@ -134,6 +140,23 @@ func (e ExecutionData) truncateTable(ctx context.Context, client ClientMeta, par
...
 }
 
+// cleanupStaleData cleans resources in table that weren't update in the latest table resolve execution
+func (e ExecutionData) cleanupStaleData(ctx context.Context, client ClientMeta, parent *Resource) error {
+	// Only clean top level tables
+	if parent != nil {
+		return nil
+	}
+	if !e.disableDelete {
+		client.Logger().Debug("skipping stale data removal", "table", e.Table.Name)
```
yevgenypats, line 150: I suggest putting this in info just like line 153 is also info (or put 153 in debug).

yevgenypats (changes requested):
some nit-picks.

yevgenypats (approved):

PR #104, opened 2021-10-18
feat/sdk execution diagnostics

roneli:


cqproto/internal/plugin.proto
```diff
@@ -70,11 +92,11 @@ message PartialFetchFailedResource {
 
 message Diagnostic {
   enum Type {
-      Unknown = 0;
-      RESOLVING = 1;
-      ACCESS = 2;
-      THROTTLE = 3;
-      DATABASE = 4;
+    Unknown = 0;
```
michelvocks, line 95: NIT: Should this be uppercase as well like the others?

cqproto/provider.go
```diff
@@ -111,6 +111,31 @@ type FetchResourcesResponse struct {
 	Error string
 	// list of resources where the fetching failed
 	PartialFetchFailedResources []*FailedResourceFetch
+	//
```
michelvocks, line 114: NIT: Missing comment

PR #105, opened 2021-10-19
Fix drop provider tables due to out of shared memory

spangenberg:
A large number of tables exceeds the transaction memory limit of
usual database configurations.


roneli (approved):
LGTM

PR #107, opened 2021-11-02
Support passing table meta information over cqproto

roneli:


cqproto/provider.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 
 	"github.com/cloudquery/cq-provider-sdk/provider/schema"
+
```
disq, line 8: stray newline

provider/schema/column.go
```diff
@@ -275,3 +278,38 @@ func (c Column) checkType(v interface{}) bool {
 
 	return false
 }
+
+func (c Column) GetMeta() *ColumnMeta {
```
disq, line 282: I would call this `Meta()` to make it [idiomatic](https://golang.org/doc/effective_go#Getters).

cqproto/grpc.go
```diff
@@ -195,11 +195,11 @@ func tablesFromProto(in map[string]*internal.Table) map[string]*schema.Table {
 func tableFromProto(v *internal.Table) *schema.Table {
 	cols := make([]schema.Column, len(v.GetColumns()))
 	for i, c := range v.GetColumns() {
-		cols[i] = schema.Column{
+		cols[i] = schema.SetColumnMeta(schema.Column{
```
disq, line 198: Looks convoluted,  maybe there's a simpler/cleaner way to convert between `internal.Column` and `schema.Column`? a method on `intenral.Column` that gives you `schema.Column` maybe? If import cycles won't be an issue that is

disq (changes requested):
LGTM overall, with recommendations

cqproto/grpc.go
roneli: internal.Column is auto generated

disq (approved):

PR #115, opened 2021-11-23
Add column name to resolver errors

### LINKED ISSUE #114: resolve columns error: invalid CIDR address: --- Test FAIL:
In tests and I guess in general it's very hard to debug when resolvers fail.

For example the following error doesn't indiciate what column and what it contains:

```
resolve columns error: invalid CIDR address: --- FAIL: TestEc2Eips (0.04s)
```

spangenberg:
Fixes #114

It's not easy to get the value in general but the column name is possible.


roneli (approved):
LGTM

PR #122, opened 2021-12-03
unit test verification now reports which columns are NULL

### LINKED ISSUE #111: resource.go:157: row at table has an empty column
Currently when testing you can get the following error:

```
    resource.go:157: row at table aws_ec2_ebs_snapshots has an empty column
```

It's very hard to debug this without either knowing which column is empty or at least printing the violating row to debug this manually.

amanenk:
fixes #111

### DIFF
--- a/provider/testing/resource.go
+++ b/provider/testing/resource.go
@@ -81,7 +81,6 @@ func TestResource(t *testing.T, providerCreator func() *provider.Provider, resou
 	assert.Nil(t, err)
 	sequence := getVerificationSequence(resource.Table)
 	verifyColumnsBySequence(t, conn, sequence, nil)
-	//verifyNoEmptyColumns(t, resource, conn)
 }
 
 func findResourceFromTableName(table *schema.Table, tables map[string]*schema.Table) string {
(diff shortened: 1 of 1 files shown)


provider/testing/resource.go
```diff
@@ -126,44 +128,71 @@ func getEnv(key, fallback string) string {
...
 	return res
 }
+func verifyColumnsBySequence(t *testing.T, conn pgxscan.Querier, sequence VerificationSequence, parentID *string) {
+	sq := squirrel.StatementBuilder.
+		PlaceholderFormat(squirrel.Dollar).
+		Select(fmt.Sprintf("json_agg(%s)", sequence.TableName)).
+		From(sequence.TableName)
+	if parentID == nil {
+		// it is a root object
+		sq = sq.Where(squirrel.Eq{"account_id": "testAccount"})
```
roneli, line 158: this seems AWS specific, how would this work on other providers?

roneli (changes requested):
I think the new code is not generic enough, needs some rethinking, imo the SQL query should be more verbose in it's return value.

provider/testing/resource.go
amanenk: Yeah. That is a great suggestion.

roneli:
Fixed in #137

PR #126, opened 2021-12-09
Add Warning About File Descriptor Usage

bbernays:
fixes: https://github.com/cloudquery/cloudquery/issues/285

### DIFF
--- a/provider/schema/execution.go
+++ b/provider/schema/execution.go
@@ -68,10 +68,13 @@ func (p ResourceFetchError) Error() string {
 	return p.Err.Error()
 }
 
-// partialFetchFailureBufferLength defines the buffer length for the partialFetchChan.
-const partialFetchFailureBufferLength = 10
+const (
+	// partialFetchFailureBufferLength defines the buffer length for the partialFetchChan.
+	partialFetchFailureBufferLength = 10
 
-const fdLimitMessage = "try increasing number of available file descriptors via `ulimit -n 10240` or by increasing timeout via provider specific parameters"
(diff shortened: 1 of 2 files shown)


yevgenypats (approved):

provider/schema/execution.go
```diff
@@ -197,6 +198,9 @@ func (e ExecutionData) callTableResolve(ctx context.Context, client ClientMeta,
 		err := e.Table.Resolver(ctx, client, parent, res)
 		if err != nil && e.Table.IgnoreError != nil && e.Table.IgnoreError(err) {
 			client.Logger().Warn("ignored an error", "err", err, "table", e.Table.Name)
+			if strings.Contains(err.Error(), ": socket: too many open files") {
+				client.Logger().Warn("try increasing number of available file descriptors via `ulimit -n 10240` or by increasing timeout via provider specific parameters")
+			}
```
roneli, line 203: The problem with this error which is okay, its not enough, the user will need to dig in the log to find the warning. Also this code will only enter if it is ignored. I suggest restructuring the if section, also setting the error with `diag.FromError` as a a Throttle diagnostic with severity warning and description etc' so user can fix this

roneli (changes requested):
Good start, but we can improve this by returning a Diagnostic to the user.

provider/schema/execution.go
bbernays: Ok! I think I implemented what you described. Let me know if it is not what you are looking for

roneli (approved):
LGTM, minor nitpicks

PR #135, opened 2021-12-22
added column name length check to unit test

amanenk:
closes cloudquery/cloudquery-issues#190


disq:
Already fixed in #133 and #134

amanenk:
closing since it is already implemented

PR #137, opened 2021-12-22
snapshot testing sdk

yevgenypats:
This is still wip and I suggest merging this after our tests are back to green so I can validate this doesn't ruin anything.

Basically it remove a lot of complex logic and instead of copying the necessary terraform files for every single test just use `-target` https://learn.hashicorp.com/tutorials/terraform/resource-targeting?in=terraform/cli

This should make the tests more robust and it's preliminary to other improvements (like snapshot testing) we want to make in the future to make our tests more developer friendly and robust.


amanenk (commented):
It is a good idea. We need to think about right organization of resources to ease targets configuration. `target` argument uses specific resource or module so we need to move all files to separate modules and use module names as parameters or we can try to list all the resources from provided files and pass them as a targets. I like the case with modules more.

provider/testing/resource.go
```diff
@@ -3,134 +3,260 @@ package testing
...
 
-	testProvider := providerCreator()
+	if err := deleteTables(conn, resource.Table); err != nil {
+		t.Fatal(err)
+	}
 
-	// No need for configuration or db connection, get it out of the way first
-	testTableIdentifiersForProvider(t, testProvider)
+	if resource.SnapshotsDir != "" {
+		if err = fetch(t, &resource); err != nil {
```
amanenk, line 79: it should be executed even if we run mock tests

yevgenypats: done

PR #139, opened 2021-12-29
show missing columns in tests

yevgenypats:


provider/testing/resource.go
```diff
@@ -126,35 +131,70 @@ func deleteTables(conn *pgxpool.Conn, table *schema.Table) error {
...
+		if len(nilColumnsArr) != 0 {
+			b, err := json.MarshalIndent(data, "", "\t")
+			if err != nil {
+				t.Fatal(err)
+			}
+			t.Errorf("found nil column in table %s. rows=\n%s\ncolumns=%s\n", table.Name, string(b), strings.Join(nilColumnsArr, ","))
 		}
+		// if tc.SkipEmptyJsonB {
+		// 	continue
+		// }
```
roneli, line 196: remove?

provider/schema/table.go
```diff
@@ -42,6 +42,10 @@ type Table struct {
 	// AlwaysDelete will always delete table data on fetch regardless if delete is disabled on run,
 	// use this only in specific cases, if you are unsure contact the CloudQuery Team.
 	AlwaysDelete bool
+	// IgnoreInTests if true this skips this column in tests as sometimes it might be hard
+	// to create a reproducible test environment with this column being non nill. For example various error tables such as
+	// security violoations and so.
```
roneli, line 47:
```suggestion
	// IgnoreInTests if true this skips this column in tests as sometimes it might be hard
	// to create a reproducible test environment with this column being non nil. For example various error tables such as
	// security violations and so.
```

roneli (approved):
Some typos

provider/testing/resource.go
yevgenypats: Im not sure where it is used. I don't think it is used in aws at least. if it's used somewhere else we can put it back.

PR #148, opened 2022-01-16
fix: Remove disable data delete option

disq:
This was set to always `true` in the previous releases.


provider/schema/execution_test.go
```diff
@@ -354,17 +349,18 @@ func TestExecutionData_ResolveTable(t *testing.T) {
...
 		assert.Equal(t, expectedResource.data["name"], "other")
 		assert.Nil(t, err)
-		exec = NewExecutionData(mockDb, logger, testTable, false, nil, false)
-		mockDb.On("CopyFrom", mock.Anything, mock.Anything, false, mock.Anything).Return(nil)
+		exec = NewExecutionData(mockDb, logger, testTable, nil, false)
+		mockDb.On("CopyFrom", mock.Anything, mock.Anything, true, mock.Anything).Return(nil)
+		mockDb.On("RemoveStaleData", mock.Anything, testTable, exec.executionStart, mock.Anything).Return(nil)
 		_, err = exec.ResolveTable(context.Background(), mockedClient, nil)
-		mockDb.AssertNumberOfCalls(t, "Delete", 1)
+		mockDb.AssertNumberOfCalls(t, "RemoveStaleData", 2)
```
roneli, line 356: did you check why this was called twice? feels wrong it should only be called once

roneli (changes requested):
Had one comment regarding tests, as it raises an alarm why it was called twice although remove stale data should be called (per each table including relations). If testTable has also a relation than 2 makes sense and we can resolve this can merge.

roneli (approved):
LGTM

PR #150, opened 2022-01-18
fix: Redact DSN parse errors

disq:
pgconn does a good job of redacting the password, but it wraps the net/url's parse error, which includes the original URI
dburl returns url.Parse errors as-is


roneli (approved):
LGTM

PR #156, opened 2022-01-19
feat: Support global tables

### LINKED ISSUE #152: Support global table
Some tables are static / global meaning fetching them each time returns the same data. 
Global tables should not fail if data already exists but rather replace, i.e latest fetch takes precedent.

roneli:
feat: support global tables option on insert it will replace conflicted rows. Closes #152

### DIFF
--- a/provider/schema/table.go
+++ b/provider/schema/table.go
@@ -46,6 +46,9 @@ type Table struct {
 	// to create a reproducible test environment with this column being non nil. For example various error tables such as
 	// security violations and so.
 	IgnoreInTests bool
+	// Global tables are usually the same regardless of the provider fetch configuration. Global table data gets replaced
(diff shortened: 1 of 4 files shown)


provider/schema/table.go
```diff
@@ -46,6 +46,9 @@ type Table struct {
 	// to create a reproducible test environment with this column being non nil. For example various error tables such as
 	// security violations and so.
 	IgnoreInTests bool
+	// Global tables are usually the same regardless of the provider fetch configuration. Global table data gets replaced
```
disq, line 49:
```suggestion
	// Global tables are usually the same regardless of the provider fetch configuration. Global table data gets fetched
```

disq (approved):

PR #157, opened 2022-01-19
fix: Migrator should consider dialect columns

disq:
This will stop the warning messages when running properly
It will start adding cq_fetch_date if running on an improperly initialised (not TSDB, but in `tsdb://` mode) DB though

### DIFF
--- a/migration/table.go
+++ b/migration/table.go
@@ -150,6 +150,9 @@ func (m TableCreator) DiffTable(ctx context.Context, conn *pgxpool.Conn, schemaN
 			m.log.Warn("column missing from table, not adding it", "table", t.Name, "column", d)
 			continue
 		}
+		if col.Internal() {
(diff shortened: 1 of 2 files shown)


roneli (commented):
Question,  just to clarify this solves if someone is running migrator on a timescaledb to create a postgres migration? although why should it care? you should up on that dialect only and then diff. Or did I miss something in this PR.

disq:
> Question, just to clarify this solves if someone is running migrator on a timescaledb to create a postgres migration? although why should it care? you should up on that dialect only and then diff. Or did I miss something in this PR.

Between our two same-protocol dialects and placeholder functions in migrations, there's no reason why we shouldn't support this. It would make the life of the developer easier.

roneli (approved):
LGTM

PR #158, opened 2022-01-20
fix: Allow reconfiguring provider in unmanaged mode

disq:
info level maybe too low, make it warn? not sure.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -60,6 +60,8 @@ type Provider struct {
 	extraFields map[string]interface{}
 	// storageCreator creates a database based on requested engine
 	storageCreator func(ctx context.Context, logger hclog.Logger, dbURL string) (schema.Storage, error)
+	// unmanaged tells if the provider is being run in unmanaged mode
+	unmanaged bool
 }
 
 func (p *Provider) GetProviderSchema(_ context.Context, _ *cqproto.GetProviderSchemaRequest) (*cqproto.GetProviderSchemaResponse, error) {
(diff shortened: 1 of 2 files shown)


provider/provider.go
```diff
@@ -93,12 +95,19 @@ func (p *Provider) GetProviderConfig(_ context.Context, _ *cqproto.GetProviderCo
...
 func (p *Provider) ConfigureProvider(_ context.Context, request *cqproto.ConfigureProviderRequest) (*cqproto.ConfigureProviderResponse, error) {
-	if p.meta != nil {
-		return &cqproto.ConfigureProviderResponse{Error: fmt.Sprintf("provider %s was already configured", p.Name)}, fmt.Errorf("provider %s was already configured", p.Name)
-	}
 	if p.Logger == nil {
 		return &cqproto.ConfigureProviderResponse{Error: fmt.Sprintf("provider %s logger not defined, make sure to run it with serve", p.Name)}, fmt.Errorf("provider %s logger not defined, make sure to run it with serve", p.Name)
 	}
+
+	if p.meta != nil {
+		if !p.Unmanaged() {
```
roneli, line 103: nit: why not just create a function that checks if CQ_PROVIDER_DEBUG = 1 and call it? instead of adding a new field.

roneli (changes requested):
I agree with the changes, I think that maybe we can make it a little shorter with a IsDebug function

roneli (approved):
LGTM

PR #161, opened 2022-01-24
fix: Resolve count

roneli:


provider/schema/execution.go
```diff
@@ -267,24 +268,28 @@ func (e *ExecutionData) resolveResources(ctx context.Context, meta ClientMeta, p
 	shouldCascade := parent == nil
 	var err error
 	resources, err = e.copyDataIntoDB(ctx, resources, shouldCascade)
+	totalCount := uint64(len(resources))
```
disq, line 271: `copyDataIntoDB` always returns `nil` resources if returning with an `err`.

provider/schema/execution.go
```diff
@@ -228,10 +228,11 @@ func (e ExecutionData) callTableResolve(ctx context.Context, client ClientMeta,
 		if len(objects) == 0 {
 			continue
 		}
-		if err := e.resolveResources(ctx, client, parent, objects); err != nil {
+		resolvedCount, err := e.resolveResources(ctx, client, parent, objects)
+		if err != nil {
 			return 0, err
 		}
-		nc += uint64(len(objects))
+		nc += resolvedCount
```
disq, line 235: `insertCount`? not sure. the method is not called `insertResources` but `resolve` is something else (calling resolvers and gathering data) and not really related to data persistence?

roneli: its total resources that were fully resolved i.e fetched + inserted.

provider/schema/execution.go
disq: so why not move totalCount lengh thing after the iferr? if err != nil, totalCount will always be zero, so why not do return 0, err below?

roneli: correct fixed

disq (approved):

PR #162, opened 2022-01-24
fix: Resource keys in resolve columns

roneli:
Fix resources keys in resolve columns

### DIFF
--- a/provider/schema/resource.go
+++ b/provider/schema/resource.go
@@ -57,12 +57,12 @@ func (r *Resource) PrimaryKeyValues() []string {
 		// we can have more types, but PKs are usually either ints, strings or a structure
 		// hopefully supporting Stringer interface, otherwise we fallback
 		switch v := data.(type) {
+		case fmt.Stringer:
+			results = append(results, v.String())
 		case *string:
 			results = append(results, *v)
 		case *int:
 			results = append(results, fmt.Sprintf("%d", *v))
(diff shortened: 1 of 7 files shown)


provider/schema/resource.go
```diff
@@ -52,7 +52,7 @@ func (r *Resource) Keys() []string {
 	for _, primKey := range tablePrimKeys {
 		data := r.Get(primKey)
 		if data != nil {
-			results = append(results, fmt.Sprintf("%v", data))
+			results = append(results, fmt.Sprintf("%s", data))
```
disq, line 55: does print `%s` with `interface{}` type work? are we sure? do all our data have stringer, or?

provider/schema/resource.go
```diff
@@ -43,16 +43,28 @@ func NewResourceData(dialect Dialect, t *Table, parent *Resource, item interface
...
+			continue
+		}
+		// we can have more types, but PKs are usually either ints, strings or a structure
+		// hopefully supporting Stringer interface, otherwise we fallback
+		switch v := data.(type) {
+		case *string:
+			results = append(results, *v)
+		case *int:
+			results = append(results, fmt.Sprintf("%d", *v))
+		case fmt.Stringer:
```
disq, line 64: nit: doesn't really matter at runtime but `fmt.Stringer` case should be the top choice I think

disq (approved):

PR #166, opened 2022-01-28
fix: Config error message is more informative now

amanenk:
config message is more informative now. It was reporting errors about both tries of decoding config: as hcl and as json  

bad config example:
```
provider "aws" {
  configuration {
    accounts "cq" {
      local_profile = "cq"
    }
    accounts "default" {
      localProfile = "default"
    }
  }
  resources            = [
        "*"
  ]
  enable_partial_fetch = true
}
```

before:
```
failed to read config as json: config.json:1,1-9: Invalid JSON keyword; "accounts" is not a valid JSON keyword., and 1 other diagnostic(s)
```
now:

```
rpc error: code = Unknown desc = failed to read config as hcl: config.hcl:5,3-15: Unsupported argument; An argument named "localProfile" is not expected here. Did you mean "local_profile"?
failed to read config as json: config.json:1,1-9: Invalid JSON keyword; "accounts" is not a valid JSON keyword., and 1 other diagnostic(s)
```


roneli (approved):
LGTM, would move errors to the response and return nil, but this is a more general fix for SDK providers so we won't get RPC errors anymore.

provider/provider.go
```diff
@@ -125,11 +125,12 @@ func (p *Provider) ConfigureProvider(_ context.Context, request *cqproto.Configu
 	if len(request.Config) == 0 {
 		p.Logger.Info("Received empty configuration, using only defaults")
 	} else if err := hclsimple.Decode("config.hcl", request.Config, nil, providerConfig); err != nil {
+		hclErr := fmt.Errorf("failed to read config as hcl: %w", err)
 		p.Logger.Warn("Failed to read config as hcl, will try as json", "error", err)
 		// this part will be deprecated.
 		if err := hclsimple.Decode("config.json", request.Config, nil, providerConfig); err != nil {
 			p.Logger.Error("Failed to load configuration.", "error", err)
-			return &cqproto.ConfigureProviderResponse{}, err
+			return &cqproto.ConfigureProviderResponse{}, fmt.Errorf("%s\nfailed to read config as json: %w", hclErr.Error(), err)
```
roneli, line 133: would a comment here, that in case of hclErr i.e diags, you want to create a better printable version of it (as it hides diagnostic as your example)

roneli (approved):
LGTM, this will require a change in the cloudquery where ConfigureProvider is called to support reading the response and checking for error there.

bbernays:
What else do we have to do in order to merge this PR?

PR #168, opened 2022-01-28
test: Refactor TestResource function

yevgenypats:
fix: Child tables weren't deleted
test: split to subtest with t.Run for each table

Related PR: https://github.com/cloudquery/cq-provider-aws/pull/450

### DIFF
--- a/provider/testing/resource.go
+++ b/provider/testing/resource.go
@@ -2,7 +2,6 @@ package testing
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


provider/testing/resource.go
```diff
@@ -23,19 +23,19 @@ import (
...
+	// ParallelFetchingLimit limits parallel resources fetch at a time
+	ParallelFetchingLimit uint64
 }
 
 // IntegrationTest - creates resources using terraform, fetches them to db and compares with expected values
 func TestResource(t *testing.T, resource ResourceTestCase) {
-	t.Parallel()
+	if !resource.NotParallel {
+		t.Parallel()
+	}
```
bbernays, line 38: Why do you need `NotParallel` why not just check if `ParallelFetchingLimit > 1` ?

yevgenypats: Those are two different things `ParallelFetchingLimit` is just the max number of multiplexer cq-provider-sdk is spawning and `Parallel` means if the `go test` will run those tests in parallel or not. if for example we will run in parallel two integration tests for the same resources than it will raise pk errors.

provider/testing/resource.go
```diff
@@ -53,28 +53,37 @@ func TestResource(t *testing.T, resource ResourceTestCase) {
...
-	if err := deleteTables(conn, resource.Table); err != nil {
-		t.Fatal(err)
+	for _, table := range resource.Provider.ResourceMap {
+		if err := tableCreator.CreateTable(context.Background(), conn, table, nil); err != nil {
+			assert.FailNow(t, fmt.Sprintf("failed to create tables %s", table.Name), err)
+		}
+		if err := deleteTables(conn, table); err != nil {
+			t.Fatal(err)
+		}
 	}
```
bbernays, line 67: Does this delete the data or the entire table? If it is deleting the data can we call it `truncateTables`?

provider/testing/resource.go
```diff
@@ -170,37 +183,30 @@ func verifyNoEmptyColumns(t *testing.T, tc ResourceTestCase, conn pgxscan.Querie
 		}
 
 		if len(nilColumnsArr) != 0 {
-			b, err := json.MarshalIndent(data, "", "\t")
+			_, err := json.MarshalIndent(data, "", "\t")
 			if err != nil {
 				t.Fatal(err)
 			}
```
bbernays, line 189: Why are we marshalling if we aren't using the data?

roneli (approved):
LGTM

PR #169, opened 2022-01-30
chore: Remove old changelog.md

yevgenypats:
New changelog is located at: https://github.com/cloudquery/cq-provider-sdk/releases


roneli (commented):
Wouldn't remove as we loose all previous changes, would link to current change long in release in github.

PR #173, opened 2022-02-01
fix: More DSN redactions

disq:
If the DSN manages to pass through a net.Addr parser and fails there, `net.AddrError` leaks connection details in the error message.


roneli (approved):
LGTM

PR #176, opened 2022-02-01
feat: Add Support For Low level Copy

bbernays:
SDK changes to expose Low Level copy functions. used in snapshot create and restore


provider/execution/storage.go
```diff
@@ -23,4 +24,6 @@ type Storage interface {
 type QueryExecer interface {
 	pgxscan.Querier
 	Exec(ctx context.Context, query string, args ...interface{}) error
+	RawCopyTo(ctx context.Context, w io.Writer, sql string) error
+	RawCopyFrom(ctx context.Context, r io.Reader, sql string) error
```
roneli, line 28: would put this in another interface called Copier, as I don't think other Storages might support this. It is always best to keep interface as smallest as possible, so people can ask for the minimal interface they require for their function.

roneli (dismissed):

provider/execution/storage.go
bbernays: I am adding Copier to QueryExec for this implementation: https://github.com/cloudquery/cloudquery/pull/449/files

disq: Maybe leave `Copier` by itself (don't embed it somewhere else) and cast/test the passed in QueryExecer to see if it also satisfies Copier. If not, runtime error? Also make sure PgDatabase satisfies it by doing an interface check. (var _ Copier = PgDatabase ...)

roneli: You can add a new interface in the core that includes QueryExecer + Copier interfaces so anything given to it must satisfy both.

provider/execution/storage.go
```diff
@@ -24,3 +25,13 @@ type QueryExecer interface {
 	pgxscan.Querier
 	Exec(ctx context.Context, query string, args ...interface{}) error
 }
+
+type Copier interface {
+	RawCopyTo(ctx context.Context, w io.Writer, sql string) error
+	RawCopyFrom(ctx context.Context, r io.Reader, sql string) error
+}
+
+type LowLevelQueryExecer interface {
```
disq, line 34: nit: This one can be defined in core I think

PR #180, opened 2022-02-03
fix: Better counting of squashed diagnostics

disq:
This way you could squash a diag multiple times, also warning/error counts are corrected

### DIFF
--- a/provider/diag/squash.go
+++ b/provider/diag/squash.go
@@ -11,6 +11,10 @@ type SquashedDiag struct {
 }
 
 func (s SquashedDiag) Description() Description {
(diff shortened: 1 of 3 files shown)


roneli (approved):
LGTM

provider/diag/squash.go
```diff
@@ -1,22 +1,67 @@
...
+		return nil
+	}
+
+	return SquashedDiag{
+		Diagnostic: r,
+		count:      s.count,
+	}
+}
+
+type Countable interface {
```
roneli, line 55: nit: Squashable? maybe not.

roneli (approved):
LGTM

PR #182, opened 2022-02-04
fix: Don't multiplexresolve if not needed

disq:


provider/execution/execution.go
```diff
@@ -66,7 +66,10 @@ func NewTableExecutor(resourceName string, db Storage, logger hclog.Logger, tabl
 // Resolve is the root function of table executor which starts an execution of a Table resolving it, and it's relations.
 func (e TableExecutor) Resolve(ctx context.Context, meta schema.ClientMeta) (uint64, diag.Diagnostics) {
 	if e.Table.Multiplex != nil {
-		return e.doMultiplexResolve(ctx, meta, nil)
+		clients := e.Table.Multiplex(meta)
+		if len(clients) > 0 {
+			return e.doMultiplexResolve(ctx, clients, nil)
+		}
```
roneli, line 72: if clients == 1 no need to multiplex either.

roneli (changes requested):
some changes, would also add a test for it in execution.

provider/execution/execution.go
disq: yes but multiplexer might've changed something in the return value?

roneli, line 72: nit:
```suggestion
		if clients := e.Table.Multiplex(meta); len(clients) > 0 {
			return e.doMultiplexResolve(ctx, clients, nil)
		}
```

roneli (approved):
LGTM

PR #184, opened 2022-02-07
feat: Add max_goroutines

yevgenypats:
This adds another option called `max_goroutines` to keep the number of goroutines at hand and not OOM when having huge amount of projects or accounts.


provider/execution/execution.go
```diff
@@ -39,10 +40,12 @@ type TableExecutor struct {
 	executionStart time.Time
 	// columns of table, this is to reduce calls to sift each time
 	columns [2]schema.ColumnList
+	// goroutinesSem to limit number of goroutines (resources fetched) concurrently
+	goroutinesSem *semaphore.Weighted
```
roneli, line 44: nit:
```suggestion
	maxGoutines *semaphore.Weighted
```

provider/execution/execution.go
```diff
@@ -39,10 +40,12 @@ type TableExecutor struct {
 	executionStart time.Time
 	// columns of table, this is to reduce calls to sift each time
 	columns [2]schema.ColumnList
+	// goroutinesSem to limit number of goroutines (resources fetched) concurrently
```
roneli, line 43: nit: limits the number of clients fetching concurrently

provider/execution/execution.go
```diff
@@ -95,20 +100,30 @@ func (e TableExecutor) doMultiplexResolve(ctx context.Context, clients []schema.
...
+	var (
+		allDiags    diag.Diagnostics
+		doneClients = 0
+	)
 	logger := clients[0].Logger()
 	logger.Debug("multiplexing client", "count", len(clients), "table", e.Table.Name)
 	defer close(diagsChan)
 	for _, client := range clients {
+		// we can only limit on a granularity of a top table otherwise we can get deadlock
+		if parent == nil {
```
roneli, line 112: no need to check parent != nil, as table relation resolvers don't get multiplexed.

provider/provider.go
```diff
@@ -170,9 +170,20 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
...
 	// limiter used to limit the amount of resources fetched concurrently
-	var limiter *semaphore.Weighted
+	var goroutinesSem *semaphore.Weighted
+	if request.MaxGoroutines == 0 {
+		// TODO: I don't like this conversion to int64
+		goroutinesSem = semaphore.NewWeighted(int64(helpers.GetMaxGoRoutines()))
+	} else {
+		// TODO: I don't like this conversion to int64
+		goroutinesSem = semaphore.NewWeighted(int64(request.MaxGoroutines))
+	}
```
roneli, line 180: would add another cases which is -1 i.e no disable would then make this a switch case.

roneli (changes requested):
Some CR comments.

provider/execution/execution.go
yevgenypats: My idea behind the naming here was that the argument is `max_goroutines` (uint64) and the semaphore is `goroutinesSem` as it can be less than max depending on the flow.

provider/provider.go
yevgenypats: it's uint64 so it can't be `-1` also if someone specify `0` it will mean it can't run so it's not a valid value as the minimum value is `1`

helpers/limits.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/pbnjay/memory"
+)
+
+const GB_IN_BYTES uint64 = 1024 * 1024 * 1024
+const GO_ROUTINES_PER_GB uint64 = 250000
+
+func GetMaxGoRoutines() uint64 {
+	totalMemory := memory.TotalMemory()
+	if totalMemory == 0 {
+		// than we assume we have 2 GB RAM
```
disq, line 13:
```suggestion
		// assume we have 2 GB RAM
```

helpers/limits.go
```diff
@@ -0,0 +1,18 @@
+package helpers
+
+import (
+	"github.com/pbnjay/memory"
+)
+
+const GB_IN_BYTES uint64 = 1024 * 1024 * 1024
+const GO_ROUTINES_PER_GB uint64 = 250000
+
+func GetMaxGoRoutines() uint64 {
```
disq, line 10: Remove TODOs about int64 conversions, make this one return int64
If it's > uint32 cap it to maxint64

disq (changes requested):
LGTM overall with comments

helpers/limits.go
yevgenypats: Fixed it. `GetMaxGoRoutines` should return uint64 as this is what the memory can return. (it can't have negative values). The check that was missing is near the passing to the semaphore (as semaphore can have negative value) and thus we need to check that the value is not more then `MAX_INT64` and if yes just cap it at MAX_INT64 but this function should be left as is I believe.

disq: Yes but blindly casting uint64's tied to memory and multipliers to int64 won't work: https://go.dev/play/p/KtA-WjWeFcQ

disq (changes requested):

yevgenypats:
@disq I didn't blindly cast to int64 - https://github.com/cloudquery/cq-provider-sdk/pull/184/files#diff-0e3b5cd2e6abf4633712dbd2d69307b7fbd75bc71844e34306afdcd9840dab64R182 (maybe I didn't push right away as there was a conflict in my local git)

disq (approved):

PR #187, opened 2022-02-10
bug: Potentially fix deadlock

yevgenypats:


provider/provider.go
```diff
@@ -202,7 +205,9 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
 		l.Unlock()
 		if parallelResourceSem != nil {
 			if err := parallelResourceSem.Acquire(ctx, 1); err != nil {
-				return err
+				p.Logger.Error("cloudn't acquire parallelResourceSem", "provider", p.Name, "resource", resource)
```
disq, line 208:
```suggestion
				p.Logger.Error("could not acquire parallelResourceSem", "provider", p.Name, "resource", resource)
```

provider/execution/execution.go
```diff
@@ -107,11 +107,14 @@ func (e TableExecutor) doMultiplexResolve(ctx context.Context, clients []schema.
 	logger := clients[0].Logger()
 	logger.Debug("multiplexing client", "count", len(clients), "table", e.Table.Name)
 	defer close(diagsChan)
+	numberOfClients := 0
 	for _, client := range clients {
 		// we can only limit on a granularity of a top table otherwise we can get deadlock
 		if err := e.goroutinesSem.Acquire(ctx, 1); err != nil {
-			return totalResources, allDiags.Add(ClassifyError(err, diag.WithResourceName(e.ResourceName)))
+			allDiags = allDiags.Add(ClassifyError(err, diag.WithResourceName(e.ResourceName)))
+			break
```
roneli, line 115: you want to continue here, if you break you stop creating clients, unless this is intentional. 
if it is intentional I would add a log message that we stopped adding more clients and a better diagnostic that we reached MAX goroutines and details that you should increase them.

provider/execution/execution.go
```diff
@@ -107,11 +107,14 @@ func (e TableExecutor) doMultiplexResolve(ctx context.Context, clients []schema.
 	logger := clients[0].Logger()
 	logger.Debug("multiplexing client", "count", len(clients), "table", e.Table.Name)
 	defer close(diagsChan)
+	numberOfClients := 0
```
roneli, line 110: I initially used a wait group, maybe its smarter, but on the other hand adding another sync primitive is unwise.

provider/provider.go
```diff
@@ -236,7 +241,12 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
 			return nil
 		})
 	}
-	return g.Wait()
+	if err := g.Wait(); err != nil {
+		// can we wrap multiple errors here or should we just join them?
+		return err
+	}
```
roneli, line 247: no you cant you can use go-multierror and append to it but you would have to do a lock on it. As errgroup returns on first error.

provider/provider.go
```diff
@@ -188,10 +188,13 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
 	finishedResources := make(map[string]bool, len(resources))
 	l := &sync.Mutex{}
 	var totalResourceCount uint64 = 0
+	var lastErr error
```
roneli, line 191: maybe use go-multierror here instead to add more errors.

roneli (changes requested):
Some review comments. Good catches for possible deadlocks.

PR #188, opened 2022-02-11
feat: Modules V2 api

disq:
Would this be enough? Discuss.


cqproto/internal/plugin.proto
```diff
@@ -140,6 +142,19 @@ message GetProviderConfig {
...
 
+message GetProviderModuleInfo {
+  message Request {
+    string module = 1;
+    repeated uint32 preferred_versions = 2;
+  }
+  message Response {
+    uint32 version = 1;
+    map<string, bytes> info = 2;
+    repeated uint32 other_versions = 3;
```
roneli, line 153: nit: supported_versions ?

cqproto/provider.go
```diff
@@ -42,6 +45,9 @@ type CQProviderServer interface {
 	// FetchResources is called when CloudQuery requests to fetch one or more resources from the provider.
 	// The provider reports back status updates on the resources fetching progress.
 	FetchResources(context.Context, *FetchResourcesRequest, FetchResourcesSender) error
+
+	// Gets info about specific module config embedded inside provider
+	GetProviderModuleInfo(context.Context, *GetProviderModuleInfoRequest) (*GetProviderModuleInfoResponse, error)
```
roneli, line 50: Nit: GetModuleInfo

provider/module/module.go
```diff
@@ -0,0 +1,20 @@
...
+
+// InfoResponse is what the provider returns from an InfoReader request.
+type InfoResponse struct {
+	// Version of the supplied "info"
+	Version uint32
+	// Info, in the given version
+	Info map[string][]byte
+
+	// Other versions supported by the provider, if any
+	OtherVersions []uint32
```
roneli, line 19: why the space?

provider/module/module.go
```diff
@@ -0,0 +1,20 @@
+package module
+
+import (
+	"github.com/cloudquery/cq-provider-sdk/provider/diag"
+	"github.com/hashicorp/go-hclog"
+)
+
+// InfoReader is called when the user executes a module, to get provider supported metadata about the given module
+type InfoReader func(logger hclog.Logger, module string, prefferedVersions []uint32) (resp InfoResponse, diags diag.Diagnostics)
```
roneli, line 9: maybe define a default "embedded" file module info reader? As I think most providers will just write configuration / multiple configuration files.

roneli (changes requested):
Some initial comments.

provider/module/module.go
```diff
@@ -0,0 +1,65 @@
...
+				continue
+			}
+			vInt, err := strconv.ParseUint(strings.TrimSuffix(f.Name(), ".hcl"), 10, 32)
+			if err != nil {
+				continue
+			}
+			resp.SupportedVersions = append(resp.SupportedVersions, uint32(vInt))
+		}
+
+		return resp, nil
```
roneli, line 63: What if directory is split into multiple files i.e we split the drift.hcl into multiple files per service etc'? might require a change in the protocol, basically support returning a list of files instead of one big one.

provider/module/module.go
```diff
@@ -0,0 +1,65 @@
...
+	Info map[string][]byte
+	// All versions supported by the provider, if any
+	SupportedVersions []uint32
+}
+
+// Serve returns an InfoReader handler given a "moduleData" filesystem.
+// The fs should have all the required files for the modules in a moduledata/ directory, as one subdirectory per module ID.
+// Filenames are the version IDs with an .hcl extension added.
+func Serve(moduleData embed.FS) InfoReader {
+	return func(logger hclog.Logger, module string, prefferedVersions []uint32) (resp InfoResponse, err error) {
```
roneli, line 29: I am not a fan of this implementation, and hiding the response var initialization the return value.

For one I would read the embdded module info before returning the function creating a mapping of module -> versions.  Would make the check easier in the function i.e only check if exists.

If it does exists read all files (also recursively in directory) returning all files as the version configuration files.

roneli (changes requested):
Added some change requests, also probably a change in the protocol to support multiple files in the module info.

provider/module/module.go
```diff
@@ -0,0 +1,65 @@
...
+	// Info, in the given version
+	Info map[string][]byte
+	// All versions supported by the provider, if any
+	SupportedVersions []uint32
+}
+
+// Serve returns an InfoReader handler given a "moduleData" filesystem.
+// The fs should have all the required files for the modules in a moduledata/ directory, as one subdirectory per module ID.
+// Filenames are the version IDs with an .hcl extension added.
+func Serve(moduleData embed.FS) InfoReader {
```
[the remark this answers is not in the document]
disq: done, without the typo though 🙄

provider/module/module.go
```diff
@@ -0,0 +1,65 @@
...
+	"strings"
+
+	"github.com/hashicorp/go-hclog"
+)
+
+// InfoReader is called when the user executes a module, to get provider supported metadata about the given module
+type InfoReader func(logger hclog.Logger, module string, prefferedVersions []uint32) (resp InfoResponse, err error)
+
+// InfoResponse is what the provider returns from an InfoReader request.
+type InfoResponse struct {
```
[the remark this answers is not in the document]
disq: `Info`. module.Reader wouldn't work for the same reason we're not doing `Serve` (not an io.Reader)

provider/module/module.go
```diff
@@ -0,0 +1,65 @@
...
+	SupportedVersions []uint32
+}
+
+// Serve returns an InfoReader handler given a "moduleData" filesystem.
+// The fs should have all the required files for the modules in a moduledata/ directory, as one subdirectory per module ID.
+// Filenames are the version IDs with an .hcl extension added.
+func Serve(moduleData embed.FS) InfoReader {
+	return func(logger hclog.Logger, module string, prefferedVersions []uint32) (resp InfoResponse, err error) {
+		for _, v := range prefferedVersions {
+			fn := fmt.Sprintf("moduledata/%s/%v.hcl", module, v)
```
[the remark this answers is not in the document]
disq: done

provider/module/module.go
```diff
@@ -0,0 +1,65 @@
...
+			data, err := moduleData.ReadFile(fn)
+			if err != nil {
+				continue
+			}
```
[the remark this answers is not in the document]
disq: already have a warning level log below, no need for another one. we're iterating multiple prefferred versions.

provider/module/module.go
disq: I don't think we could do recursively. Maybe in next version. swapped the order.

provider/module/module.go
disq: done

roneli (approved):
LGTM

PR #192, opened 2022-02-23
test: Fix test failures when local postgres instance has a different table schema

shimonp21:
Fixes an issue where tests failed if the local postgresql had a different schema.

For example, if I added a column to a table (but postgresql still has the old table schema),
the tests would fail (because we ran "CREATE TABLE IF NOT EXISTS" and "DELETE FROM", but at no point did we actually add the new column).

It's better to drop and recreate them entirely, I believe.

### DIFF
--- a/provider/testing/resource.go
+++ b/provider/testing/resource.go
@@ -58,11 +58,8 @@ func TestResource(t *testing.T, resource ResourceTestCase) {
 	tableCreator := migration.NewTableCreator(l, schema.PostgresDialect{})
 
 	for _, table := range resource.Provider.ResourceMap {
-		if err := tableCreator.CreateTable(context.Background(), conn, table, nil); err != nil {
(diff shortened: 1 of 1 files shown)


migration/table.go
```diff
@@ -52,6 +52,29 @@ func (m TableCreator) CreateTable(ctx context.Context, conn execution.QueryExece
 	return nil
 }
 
+// DropAndCreateTable generates DROP TABLE and CREATE TABLE definitions for the given table, and runs them on the given conn
+func (m TableCreator) DropAndCreateTable(ctx context.Context, conn execution.QueryExecer, t, parent *schema.Table) error {
```
disq, line 56: We're removing `TableCreator.CreateTable` in #190, I don't think we should add a new one here (that's also accessible from the outside) so I suggest moving this into an unexported fn under `provider/testing/`.

disq (changes requested):

disq (approved):
LGTM

PR #195, opened 2022-02-28
feat: Testing: Require cq_fetch_date on every tsdb table

disq:


migration/testbuilder.go
```diff
@@ -10,12 +10,22 @@ import (
...
+	"github.com/georgysavva/scany/pgxscan"
 	"github.com/golang-migrate/migrate/v4"
 	"github.com/hashicorp/go-hclog"
+	"github.com/jackc/pgx/v4/pgxpool"
 	"github.com/spf13/afero"
 	"github.com/stretchr/testify/assert"
 )
 
+const queryPrimaryKeys = `SELECT kcu.table_name, c.constraint_name, ARRAY_AGG(kcu.column_name::text ORDER BY kcu.ordinal_position) AS pk_cols
+FROM information_schema.table_constraints c
```
yevgenypats, line 22: Can you add a bit of docs on what this query is doing ? :)

yevgenypats (approved):
Looks good but I'd add more docs as not sure what is the purpose of the query and what exactly it is checking.

migration/testbuilder.go
disq: "documented"

disq:
also, the test name (and the fn that runs the test) is self-documenting with the names.

PR #196, opened 2022-03-01
fix: Single insert

disq:
Inserting singly starts failing with bad column quotation after first operation.


bbernays (approved):
LGTM. But for my own edification can you provide an example of where this was failing?

disq:
> LGTM. But for my own edification can you provide an example of where this was failing?

Insert is used when bulk loading data doesn't work (because of PK collisions due to a broken resource etc)
Rows are iterated and inserted one by one. After the first insert, following inserts don't work and fail with a different message (which is a syntax error, due to overquoting of column names)

PR #197, opened 2022-03-01
feat: Add migrator prehook

disq:
This was needed to be able to 'drop views' on timescaledb dialect executor before the migrator was run.


bbernays (approved):
LGTM

PR #200, opened 2022-03-08
feat: Separate metadata per fetch request

disq:
Relates to [#520](https://github.com/cloudquery/cloudquery/issues/520)


provider/execution/execution.go
```diff
@@ -34,8 +34,10 @@ type TableExecutor struct {
 	Logger hclog.Logger
 	// classifiers
 	classifiers []ErrorClassifier
-	// extraFields to be passed to each created resource in the execution
+	// extraFields to be passed to each created resource in the execution, used in tests.
 	extraFields map[string]interface{}
```
roneli, line 38: I think the documentation is wrong here, this is used to add new columns, and its used in history not tests (maybe also in tests)

roneli (approved):
LGTM had one comment, on description

PR #201, opened 2022-03-09
feat: Tool: Handle PK changes on migrations

disq:
~the `DefaultSchema()` thing is bothering me.~

### DIFF
--- a/migration/table.go
+++ b/migration/table.go
@@ -3,6 +3,7 @@ package migration
 import (
 	"context"
 	"fmt"
+	"reflect"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


migration/table.go
```diff
@@ -242,3 +257,86 @@ func findSimilarColumnsWithSameType(setA, setB map[string][]string) map[string]s
...
+	for _, m := range c {
+		ret.additions = append(ret.additions, m.additions...)
+		ret.removals = append(ret.removals, m.removals...)
+	}
+	return ret
+}
+
+func (m TableCreator) diffConstraints(ctx context.Context, conn *pgxpool.Conn, schemaName string, t *schema.Table) (up, down constraintMigrations, err error) {
+	if ds, ok := m.dialect.(interface{ DefaultSchema() string }); ok {
+		schemaName = ds.DefaultSchema()
```
disq, line 279: ^ bothers me. `history` was only in cq-core until this point.

provider/schema/dialect.go
```diff
@@ -197,6 +197,10 @@ func (d TSDBDialect) GetResourceValues(r *Resource) ([]interface{}, error) {
 	return doResourceValues(d, r)
 }
 
+func (d TSDBDialect) DefaultSchema() string {
+	return "history"
```
disq, line 201: ^ more botherage

roneli: why do you need to add the default schema here? when running the migrations should be in the correct search path and they should be schema agnostic. Or maybe I missed something.

roneli (commented):
LGTM, added a comment, on the bothering parts

provider/schema/dialect.go
disq: @roneli because on TSDB the actual constraint isn't in the search path (`public`) as we have views in there. So list of PKs return 0 rows when queried with `tavle_schema='public'`

roneli (approved):

PR #202, opened 2022-03-10
feat: Add WrapError to add stacktrace

yevgenypats:


helpers/errors.go
```diff
@@ -0,0 +1,19 @@
...
+
+// WrapError wraps error with the following string: "error in function_name[filename:line_number]: %w"
+// if err is nil returns nil
+func WrapError(err error) error {
+	if err != nil {
+		// notice that we're using 1, so it will actually log the where
+		// the error happened, 0 = this function, we don't want that.
+		pc, filename, line, _ := runtime.Caller(1)
+
+		return fmt.Errorf("error in %s[%s:%d] %v", runtime.FuncForPC(pc).Name(), filename, line, err)
```
disq, line 16:
```suggestion
		return fmt.Errorf("error in %s[%s:%d] %w", runtime.FuncForPC(pc).Name(), filename, line, err)
```

disq (changes requested):
Not using %w would break most classifications.

helpers/errors.go
```diff
@@ -0,0 +1,19 @@
+package helpers
```
roneli, line 1: why not move this into diag package?

helpers/errors.go
roneli, line 16:
```suggestion
		return fmt.Errorf("error at %s[%s:%d] %v", runtime.FuncForPC(pc).Name(), filename, line, err)
```

roneli (approved):
LGTM, had to minor comments + add a test to verify this.

helpers/errors.go
```diff
@@ -0,0 +1,19 @@
...
+
+// WrapError wraps error with the following string: "error at function_name[filename:line_number]: %w"
+// if err is nil returns nil
+func WrapError(err error) error {
+	if err != nil {
+		// notice that we're using 1, so it will actually log the where
+		// the error happened, 0 = this function, we don't want that.
+		pc, filename, line, _ := runtime.Caller(1)
+
+		return fmt.Errorf("error at %s[%s:%d] %w", runtime.FuncForPC(pc).Name(), filename, line, err)
```
disq, line 16: `filename` is leaking compile path, any way to remove the prefix?

disq: Maybe removing filename can be done using going up until we hit a path where it's not a `cloudquery` or `cq-*` package... My thinking is, if there was a better way it would've been implemented [here](https://github.com/pkg/errors/blob/v0.9.1/stack.go#L172) but it's not

yevgenypats: Good catch. I added `path.Basename`

spangenberg (approved):

roneli (changes requested):
Still missing one comment to resolve

PR #204, opened 2022-03-14
fix: Tool: Add param to run in TSDB mode without TSDB

disq:


disq:
Agreed, will add after it's settled (so, now probably)

roneli (approved):

bbernays (approved):

PR #206, opened 2022-03-16
fix: Do not ignore in test for mock tests

zagronitay:


roneli (approved):
LGTM

provider/testing/resource.go
```diff
@@ -30,6 +30,8 @@ type ResourceTestCase struct {
 	NotParallel bool
 	// ParallelFetchingLimit limits parallel resources fetch at a time
 	ParallelFetchingLimit uint64
+	// IgnoreInTest detects if tests should be ignored
+	IgnoreInTest bool
```
roneli, line 34:
```suggestion
	// Skip IgnoreInTest flag which detects if scema.Table or schema.Column should be ignored
	SkipIgnoreInTest bool
```

PR #210, opened 2022-03-18
feat: Diag: Keep accountID as a field

disq:
Since `client.Meta` is very broad the account ID should be added in the error classifier of each provider, with `WithAccountID(string)`

If we were to add `AccountID() string` to the `client.Meta` interface, it can be done completely inside the SDK.


bbernays:
I feel like there might be other none sensitive metadata that we might want to retain the in the future. What if we were make a generic `map[string]string` object and allow the provider/client to define whatever it wants. Also there are some providers that do not have an "accountID" like K8s

disq:
I kind of agree but if it's just a handful of provider types I don't know. (map[string][]string would be better to accomodate resourceID as well btw)
And we can provide special strings as keys `const diag.ResourceID = "resource_id"` to make it look nice/special-formattable.

bbernays:
The other piece of information that is missing (for AWS at least) is the region, which a map would be able to support as well

disq:
> The other piece of information that is missing (for AWS at least) is the region, which a map would be able to support as well

Putting the region there or any other "spammy" ID (like if it's erroring in all regions then we'll have 26 of them) would really reduce the way we squash these diagnostics and it would lead to a river of errors. Maybe not include any info from the map in the squashing and squash them all together etc? Not sure.

disq:
> Maybe not include any info from the map in the squashing and squash them all together etc? Not sure.

But I would care about "list of accounts that are failing" but I wouldn't care about the regions... not including them in the squashing process would just give me the first account ID (and first region) which is not what I want.

bbernays:
Why do you not care about the region? Some functionality is not available in every region so including it would definitely help customers investigate any issues.  Also some customers implement different IAM permissions for different regions.

disq:
> Why do you not care about the region? Some functionality is not available in every region so including it would definitely help customers investigate any issues. Also some customers implement different IAM permissions for different regions.

Maybe we'd need a region-aware squasher? if the resource type is erroring on all regions just squash it into one, if not, don't.

bbernays:
> Maybe we'd need a region-aware squasher? if the resource type is erroring on all regions (available to that type) just squash it into one, if not, don't.

Yeah I think that is the right balance.

roneli (commented):
The code by itself seems fine, I am not sure I agree with the implementation similar to what @bbernays said. AccountId is very specific to some providers ~, why not use a JSON field that the top level provider can add ? I am assuming this is added in the classifier, so now we added account ID but tomorrow we would like to add more information. But truthfully the the error classifier could have used the ResourceId to convey this information i.e append account/region to the resource Id.

PR #211, opened 2022-03-19
fix: IgnoreErrors were showing as warn and errors

yevgenypats:


disq:
This may not be right. Should show up as ignore, not actually ignored. Could be something else, but it was fine last time I checked.

yevgenypats:
I just checked it with latest GCP and what was happening is that even if Errors were ignored they were returned as ResolvingErrors because of this line https://github.com/cloudquery/cq-provider-sdk/blob/87a1ce9ba2917d846d85cc1420c5bea0c0bec991/provider/execution/execution.go#L408

yevgenypats:
Currently, what I did I just removed diags from ignored errors completely and just kept the log. I can add it back but I don't think it's useful as it's ignored by the provider. 
If we add it back in anycase it should be wrapped with it's own diag as currently it is wrapped with Ignore and then with Error

roneli:
> I just checked it with latest GCP and what was happening is that even if Errors were ignored they were returned as ResolvingErrors because of this line
> 
> https://github.com/cloudquery/cq-provider-sdk/blob/87a1ce9ba2917d846d85cc1420c5bea0c0bec991/provider/execution/execution.go#L408

I think the bug is here, it shouldn't override the error level by default @disq thoughts? Originally ignored errors weren't passed, but they were requested to be, the were added in IgnoreLevel. We can remove them, but I personally think we should convey them to pass maximum info.

yevgenypats:
I'm in favor of ignoring them completely (and just keeping a log-line) as this is what the function suggests. Otherwise, it really creates spammy results for the user in an ETL framework and is hard to find/understand what should be addressed and what not. 

In  Sentry we ignore all the warnings anyway because of their massive ammount as they create a lot of noise so I would really push toward ignoring this at the core.

roneli (approved):

disq:
> I think the bug is here, it shouldn't override the error level by default @disq thoughts?

@roneli `fromError`, when passed a BaseError as error, won't overwrite the errors (because opts are applied after `diag.WithNoOverwrite`) and it was working fine last time I tested with AWS. Problem seems to be in the GCP error classifier as I predicted, but there's no point now that this is merged and released.

PR #213, opened 2022-03-23
fix: Resolve column diags not returned

roneli:


disq:
Consider deprecating WithNoOverwrite with:
```
Deprecated: Prefer using WithOptionalSeverity on the opposite side instead
```

Removing WithNoOverwrite would also simplify all the `BaseOption` fn bodies.

disq (approved):

PR #214, opened 2022-03-23
fix: Clean up reattach file reading

disq:
Tests sometimes fail with "invalid json", maybe it attempts reading before the file is closed/flushed on the provider side? This might fix it.


roneli (approved):
LGTM

PR #215, opened 2022-03-24
fix: Diagnostic counts

disq:


provider/diag/diagnostics.go
```diff
@@ -126,19 +126,25 @@ func (diags Diagnostics) Squash() Diagnostics {
...
 }
 
 func (diags Diagnostics) Errors() uint64 {
-	return diags.CountBySeverity(ERROR)
+	return diags.CountBySeverity(ERROR, false)
 }
 
-func (diags Diagnostics) CountBySeverity(sev Severity) uint64 {
+// CountBySeverity returns number of diagnostics of the given severity. If deep is false, squashed diags are counted as a single diag.
+func (diags Diagnostics) CountBySeverity(sev Severity, deep bool) uint64 {
```
roneli, line 137: don't like the argument deep its ambiguous, maybe `include_squashed` or  `count_squashed` so its obvious by the param name what this flag turns on.

roneli (changes requested):
LGTM, had one nit around the param name

provider/diag/diagnostics.go
disq: fixed

roneli (approved):

PR #221, opened 2022-04-14
feat: Testing. More ways to verify table content.

daniil-ushkov:
Hi! 
In [yandex-cloud provider](https://github.com/yandex-cloud/cq-provider-yandex) we want to verify more complex conditions than just non-empty columns. So I suggest to provide opportunity of writing custom verifiers for user.

### DIFF
--- a/provider/testing/resource.go
+++ b/provider/testing/resource.go
@@ -32,7 +32,10 @@ type ResourceTestCase struct {
 	ParallelFetchingLimit uint64
 	// SkipIgnoreInTest flag which detects if schema.Table or schema.Column should be ignored
 	SkipIgnoreInTest bool
+	Verifiers        []Verifier
 }
+
+type Verifier func(t *testing.T, conn pgxscan.Querier)
 
 func init() {
 	_ = faker.SetRandomMapAndSliceMinSize(1)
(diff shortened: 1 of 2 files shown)


provider/testing/verifiers.go
```diff
@@ -0,0 +1,80 @@
+package testing
```
roneli, line 1: Would add documentation for each verifier type, and what they are checking

provider/testing/resource.go
```diff
@@ -32,8 +32,11 @@ type ResourceTestCase struct {
 	ParallelFetchingLimit uint64
 	// SkipIgnoreInTest flag which detects if schema.Table or schema.Column should be ignored
 	SkipIgnoreInTest bool
+	Verifiers        []Verifier
 }
 
+type Verifier func(t *testing.T, conn pgxscan.Querier)
```
roneli, line 38: would maybe make the verifier signature like this, and the default verifier will be no empty columns
```suggestion
type Verifier func(t *testing.T, table *schema.Table, conn pgxscan.Querier, SkipIgnoreInTest bool)
```

provider/testing/resource.go
```diff
@@ -32,8 +32,11 @@ type ResourceTestCase struct {
 	ParallelFetchingLimit uint64
 	// SkipIgnoreInTest flag which detects if schema.Table or schema.Column should be ignored
 	SkipIgnoreInTest bool
+	Verifiers        []Verifier
```
roneli, line 35: nit: document this field

roneli (changes requested):
Looks like a great addition to SDK testing, had a few suggestions

provider/testing/resource.go
daniil-ushkov: To support ignoring flag, I changed signature as you suggested. Now single verifier will be used to verify one or more tables (main table and relations) from specific *schema.Table. So in ResourceTestCase's Verifiers field we have map from resource name to list of verifiers for this resource.

roneli (approved):
LGTM

yevgenypats:
Hi @daniil-ushkov Looks great, thx! I'd love to run a few more potential improvement that we are thinking on the SDK side. Can you drop me a note on yp@cloudquery.io (couldn't find your email easily :) )?

shimonp21:
@daniil-ushkov Hi :) Updating that we released version [v0.8.19](https://github.com/cloudquery/cq-provider-sdk/releases/tag/v0.8.19) of the SDK.

PR #226, opened 2022-04-22
feat: ConfigureProvider to return diagnostics, add diag.USER and UnsquashDiag

disq:
Error field is left for backwards compatibility in proto, it's generated from `response.Diagnostics.Error()`
`diag.USER` is finally there for user-caused errors like bad configuration etc.


provider/provider.go
```diff
@@ -96,12 +96,16 @@ func (p *Provider) GetProviderConfig(_ context.Context, _ *cqproto.GetProviderCo
 
 func (p *Provider) ConfigureProvider(_ context.Context, request *cqproto.ConfigureProviderRequest) (*cqproto.ConfigureProviderResponse, error) {
```
disq, line 97: `error` is unused in the signature but removing it fails the CQServer interface check (which is unrelated I think...)

provider/provider.go
```diff
@@ -96,12 +96,16 @@ func (p *Provider) GetProviderConfig(_ context.Context, _ *cqproto.GetProviderCo
...
 	if p.Logger == nil {
-		return &cqproto.ConfigureProviderResponse{Error: fmt.Sprintf("provider %s logger not defined, make sure to run it with serve", p.Name)}, fmt.Errorf("provider %s logger not defined, make sure to run it with serve", p.Name)
+		return &cqproto.ConfigureProviderResponse{
+			Diagnostics: diag.FromError(fmt.Errorf("provider %s logger not defined, make sure to run it with serve", p.Name), diag.USER),
```
roneli, line 100: don't this is a user related error, but more internal as in the provider dev

roneli (approved):
Had one nit.

provider/provider.go
disq: fixed

provider/diag/error.go
```diff
@@ -201,3 +201,39 @@ func FromError(err error, dt DiagnosticType, opts ...BaseErrorOption) Diagnostic
 		return Diagnostics{NewBaseError(err, dt, opts...)}
 	}
 }
+
+// Mutate generates new Diagnostics (using BaseError) from the given Diagnostics, and applies the given options to each one
+func Mutate(dd Diagnostics, opts ...BaseErrorOption) Diagnostics {
```
roneli, line 206: where is this used? its kinda cool but still

roneli (approved):
Looks good, had a question on the mutate diags method

provider/diag/error.go
disq: @roneli To distinguish between Configure diags vs. Fetch diags. See https://github.com/cloudquery/cloudquery/pull/594 and the last commit in it

roneli: I don't think you need to set a resource name / id for diagnostics as its only for "resource diags", but I see the usage

provider/diag/diagnostic.go
```diff
@@ -25,6 +25,8 @@ func (d DiagnosticType) String() string {
 		return "Throttle"
 	case DATABASE:
 		return "Database"
+	case USER:
```
roneli, line 28: I know this isn't part of the PR review, but case Internal is missing.

roneli (changes requested):
Added nit

provider/diag/diagnostic.go
disq: fixed

provider/diag/error.go
roneli: So we don't need this function basically, we can remove if not.

provider/provider.go
roneli: We should usually avoid returning error and put it in the response, as that causes the GRPC fatal which no user understands why its "fatal"

roneli (approved):
LGTM, added 2 comments

PR #229, opened 2022-04-26
fix: Backwards compatibility with old CQ Core

disq:
Better than https://github.com/cloudquery/cq-provider-sdk/pull/228


disq:
Unless there's a way to override this behaviour here, this PR won't work.
https://github.com/cloudquery/cq-provider-sdk/blob/main/cqproto/internal/plugin_grpc.pb.go#L62-L65

PR #233, opened 2022-04-27
feat: More logging around execute/multiplex

disq:


spangenberg (commented):
LGTM

provider/execution/execution.go
```diff
@@ -87,7 +87,8 @@ func (e TableExecutor) Resolve(ctx context.Context, meta schema.ClientMeta) (uin
 		ctx, cancel = context.WithTimeout(ctx, e.timeout)
 		defer cancel()
 	}
-	return e.callTableResolve(ctx, meta, nil)
+	logger := meta.Logger().With("table", e.Table.Name, "multiplex", false)
+	return e.callTableResolve(ctx, logger, meta, nil)
```
roneli, line 91: would actually do e.WithLogger(...).callTableResolver and inside execution call the logger instead, as you want the table to also change when going to sub relation or do  resource: "X" and table: "Y" where Y can change to relation.

Moreover would also add to the call the multiplex info i.e in aws case region + account_id as you can have multiple execution of a same resource but you want to distinguish them by region/account_id

provider/execution/execution.go
```diff
@@ -126,62 +127,65 @@ func (e TableExecutor) doMultiplexResolve(ctx context.Context, clients []schema.
...
 			defer wg.Done()
 			if e.timeout > 0 {
 				var cancel context.CancelFunc
 				ctx, cancel = context.WithTimeout(ctx, e.timeout)
 				defer cancel()
 			}
+			defer lgr.Debug("releasing multiplex client", "ctx_err", ctx.Err())
 
-			count, resolveDiags := e.callTableResolve(ctx, c, nil)
+			count, resolveDiags := e.callTableResolve(ctx, lgr, c, nil)
```
roneli, line 155: see my comments around With

provider/execution/execution.go
```diff
@@ -126,62 +127,65 @@ func (e TableExecutor) doMultiplexResolve(ctx context.Context, clients []schema.
...
 	for _, client := range clients {
 		// we can only limit on a granularity of a top table otherwise we can get deadlock
+		logger.Debug("trying acquire for new client", "next_id", numberOfClients+1)
 		if err := e.goroutinesSem.Acquire(ctx, 1); err != nil {
 			diagsChan <- ClassifyError(err, diag.WithResourceName(e.ResourceName))
 			break
 		}
-		logger.Debug("creating multiplex client new client")
 		numberOfClients++
+		clientLogger := logger.With("client_id", numberOfClients)
```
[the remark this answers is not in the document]
roneli: We can also have a number ID so its easier inside the log, but across different runs we want it to be consistent.

roneli (changes requested):
Some thoughts around the added logging, I want to look at the execution also to see if any more lines to suggest.

provider/execution/execution.go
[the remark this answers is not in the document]
disq: we can't have provider-specific identifiers unless we add a new method to `ClientMeta` to ask them from the provider.

provider/execution/execution.go
disq: Same comment applies here. This is the best we can do without breaking interfaces.

roneli: I think a with would be cleaner then passing logger to all methods. I would actually break yes, add to client.Meta a new interface method called Identifiers() []Identifiers-> [account_id, region], otherwise we wont be able to consistently distinguish specific breaking clients by their identifier / problematic i.e it can be a specific account / region doing issues etc'

disq: with wouldn't work with all methods as it would create new execution?

disq:
@roneli I tried the `withLogger` approach but it makes things very implicit and confusing.

disq:
or #235

PR #241, opened 2022-05-05
feat: Column resolver detailed panic info

disq:


irmatov (approved):

provider/execution/execution.go
```diff
@@ -397,10 +397,20 @@ func (e TableExecutor) resolveResourceValues(ctx context.Context, meta schema.Cl
...
+	var col string
+
+	defer func() {
+		if r := recover(); r != nil {
+			stack := string(debug.Stack())
+			e.Logger.Error("resolve columns recovered from panic", "panic_msg", r, "stack", stack, "column_name", col)
+			diags = fromError(fmt.Errorf("column resolve panic: %s", r), diag.WithResourceName(e.ResourceName), diag.WithSeverity(diag.PANIC),
+				diag.WithSummary("resolve column %q in table %q recovered from panic", col, e.Table.Name), diag.WithDetails("%s", stack))
+		}
+	}()
```
roneli, line 410: defer copies col by value, so the value is incorrect. Tried to recreate in playground

https://go.dev/play/p/8j21HYQVJ5a

Would also add  some tests here.

roneli (changes requested):
The general logic LGTM, but I think there is a problem with the defer as the col gets copied into the function by value, and setting it again won't pass the correct data, added an example in the playground, unless I misinterpreted it.  Would add tests to verify it regardless.

provider/execution/execution.go
disq: https://go.dev/play/p/dqhjbzg2hx2 better example, but correct

disq: Test added

PR #242, opened 2022-05-06
fix: Int64 to int automatic conversion added

### LINKED ISSUE #240: Allow sdk to convert from int64 to int automatically
### Describe the bug

Columns that described as `int` but have `int64` type do not converted to int automatically.

### Expected Behavior

columns described as int type should automatically converted from int64 to int

### Steps to Reproduce

create resource from structure that has in64 type and describe column type as int.

### Possible Solution

add automatic conversion for this pair of types.

### Provider and CloudQuery version

cloudquery 0.22.10, aws 0.11.1

### Additional Context

_No response_

amanenk:
closes #240


disq (approved):
LGTM

PR #250, opened 2022-05-12
feat: Detect duplicates from resolvers

bbernays:
`Diag` is probably wrong. Any guidance would be great


disq:
Why did we close this? Doesn't make sense? how?

PR #252, opened 2022-05-15
fix(deps): Update hashstructure

erezrokah:
Related to https://github.com/cloudquery/cloudquery-issues/issues/336 (internal issue).

We've seen the following error pop up from time to time:
```
failed to insert to table "aws_ec2_internet_gateway_attachments": ERROR: duplicate key value violates unique constraint "aws_ec2_internet_gateway_attachments_pk"
```

The reason is probably concurrent fetches, however the version we're using of `hashstructure` has a bug that can cause hash collisions. It was fixed in v2 https://github.com/mitchellh/hashstructure/tree/a045b665615f739329536a58c25ca6327abf1ec1#installation

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/jackc/pgerrcode v0.0.0-20201024163028-a0d42d470451
 	github.com/jackc/pgtype v1.8.1
 	github.com/jackc/pgx/v4 v4.13.0
-	github.com/mitchellh/hashstructure v1.1.0
+	github.com/mitchellh/hashstructure/v2 v2.0.2
 	github.com/modern-go/reflect2 v1.0.2
 	github.com/pbnjay/memory v0.0.0-20210728143218-7b4eea64cf58
 	github.com/spf13/afero v1.6.0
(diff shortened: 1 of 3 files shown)


roneli (approved):

disq:
Guessing V2 format isn't backwards compatible with V1 (given the same keys) so RemoveStaleData (or `on conflict update`?) will (probably) fail to detect updates and the user will have duplicate rows... probably

erezrokah:
Thanks for the added context @disq 👍  Lets revert in https://github.com/cloudquery/cq-provider-sdk/pull/254 until I can verify this change

roneli:
Doesn't really matter if you drop tables with new migration change, as any update to a newer version with newer cq_ids will not have conflicts after an update.

erezrokah:
> Doesn't really matter if you drop tables with new migration change, as any update to a newer version with newer cq_ids will not have conflicts after an update.

What do you say we revert until the new migrations change? This way we won't block the upcoming release of the SDK https://github.com/cloudquery/cq-provider-sdk/pull/246

PR #256, opened 2022-05-16
fix: Calculate goroutines with ulimit

roneli:


helpers/limit/limits.go
```diff
@@ -0,0 +1,32 @@
+package limit
+
+import (
+	"github.com/pbnjay/memory"
+)
+
+const gbInBytes int = 1024 * 1024 * 1024
```
disq, line 7: nit: use `const (` block

helpers/limit/limits.go
```diff
@@ -0,0 +1,32 @@
...
+
+func GetMaxGoRoutines() uint64 {
+	limit := calculateGoRoutines(getMemory())
+	ulimit, err := getUlimit()
+	if err != nil || ulimit == 0 {
+		return limit
+	}
+	if ulimit > limit {
+		return limit
+	}
```
bbernays, line 18: Not sure ulimit should equal goroutine maximum... As it is possible that a resolver might use multiple FDs... so maybe it should be
```suggestion
	if ulimit > limit {
		return uint64(limit*2 / 3)
	}
```

disq (approved):
lgtm with nit

bbernays (approved):

PR #262, opened 2022-05-18
feat!: Migrations removal

disq:
Supposed to go with https://github.com/cloudquery/cloudquery/pull/731


roneli (approved):
LGTM, don't have many comments, as mostly code removal so its much leaner. The signature code looks great.

PR #265, opened 2022-05-20
fix: Executor fixes

disq:
Various fixes

- Always use multiplexresolve
- Keep all values of `TableExecutor` in `with*()` functions
- Export CQ Fetch Id const
- CleanupStaleData is called only if we successfully fetched/resolved (otherwise failed resources sometimes empty the table)

More to come:
- Remove `ExtraFields` (sdk / core / cqproto) and all support for additional deletefilter columns etc. (this will break history so I reverted -- later in separate PR)
- Remove `AlwaysDelete`? (not used anywhere unless there's a private provider I'm not aware of)
- When ran with `CopyFrom` part disabled, the delete doesn't remove anything (and no on conflict) so it can't run more than once (fixed in https://github.com/cloudquery/cq-provider-sdk/pull/266)


roneli (approved):
LGTM

provider/execution/execution.go
```diff
@@ -299,8 +286,8 @@ func (e TableExecutor) resolveResources(ctx context.Context, meta schema.ClientM
 		diags     diag.Diagnostics
 	)
 
-	for _, o := range objects {
-		resource := schema.NewResourceData(e.Db.Dialect(), e.Table, parent, o, e.metadata, e.executionStart)
+	for i := range objects {
```
shimonp21, line 289: This change seems unnecessary, no?

provider/execution/execution.go
```diff
@@ -286,9 +269,13 @@ func (e TableExecutor) callTableResolve(ctx context.Context, client schema.Clien
 	if parent == nil {
 		e.Logger.Info("fetched successfully", "count", nc)
 	}
-	if err := e.cleanupStaleData(ctx, client, parent); err != nil {
-		return nc, diags.Add(fromError(err, diag.WithType(diag.DATABASE), diag.WithSummary("failed to cleanup stale data on table %q", e.Table.Name)))
+
+	if !diags.HasErrors() {
```
shimonp21, line 273: I have a strong gut feeling we should avoid this change for now:

- Even a single column (e.g. in a child relation) that failed to be resolved can cause "error" diags, right? So a single column with a null value in any child relation will cause failure of cleanup .

Seems dangerous in any case. I'd really avoid doing this unless we have a specific reason to do it (e.g. customers complining, sentry errors, ....).

shimonp21 (changes requested):

provider/execution/execution.go
disq: If the `interface{}` isn't a pointer then this change is necessary (and more future proof) due to the way `for` works in go.

provider/execution/execution.go
disq: if the column resolver returns error diags, it doesn't continue to complete all other tasks so (if there are any) the new rows are incomplete anyway. no postresourceresolver called, no internal resolvers (cq_id!) called. https://github.com/cloudquery/cq-provider-sdk/blob/main/provider/execution/execution.go#L367-L370

not sure if it will lead to more rows or not on edge cases.

PR #266, opened 2022-05-20
fix: Delete by cq_id before insertion

disq:
https://github.com/cloudquery/cq-provider-sdk/pull/265 Continued

I think current code assumes `CopyFrom` trial will delete all dupe ids, but it’s in a TX so as soon as it fails everything is reverted back to cause us more problems

### DIFF
--- a/database/postgres/pgdatabase.go
+++ b/database/postgres/pgdatabase.go
@@ -45,10 +45,17 @@ func NewPgDatabase(ctx context.Context, logger hclog.Logger, dsn string, sd sche
 var _ execution.Storage = (*PgDatabase)(nil)
 
 // Insert inserts all resources to given table, table and resources are assumed from same table.
-func (p PgDatabase) Insert(ctx context.Context, t *schema.Table, resources schema.Resources) error {
+func (p PgDatabase) Insert(ctx context.Context, t *schema.Table, resources schema.Resources, shouldCascade bool, cascadeDeleteFilters map[string]interface{}) error {
 	if len(resources) == 0 {
 		return nil
 	}
+
+	if shouldCascade {
+		if err := deleteResourceByCQId(ctx, p.pool, resources, cascadeDeleteFilters); err != nil {
+			return err
+		}
+	}
+
 	// It is safe to assume that all resources have the same columns
 	cols := quoteColumns(resources.ColumnNames())
 	psql := sq.StatementBuilder.PlaceholderFormat(sq.Dollar)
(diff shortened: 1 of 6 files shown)


database/postgres/pgdatabase.go
```diff
@@ -45,10 +45,17 @@ func NewPgDatabase(ctx context.Context, logger hclog.Logger, dsn string, sd sche
...
 
 // Insert inserts all resources to given table, table and resources are assumed from same table.
-func (p PgDatabase) Insert(ctx context.Context, t *schema.Table, resources schema.Resources) error {
+func (p PgDatabase) Insert(ctx context.Context, t *schema.Table, resources schema.Resources, shouldCascade bool, cascadeDeleteFilters map[string]interface{}) error {
 	if len(resources) == 0 {
 		return nil
 	}
+
+	if shouldCascade {
+		if err := deleteResourceByCQId(ctx, p.pool, resources, cascadeDeleteFilters); err != nil {
```
disq, line 54: As an alternative we could enable `ON CONFLICT UPDATE`, but that checks PK collisions (which we want to keep/see/reported) and not necessarily a `cq_id` collision.

Or running this + the insert below in a TX might be a (costly?) option.

shimonp21: I think one of the above needs to be the solution, due to the fact that when data is deleted during fetch, can cause bad user-experience, even if it is then re-inserted again anyway (@yevgenypats @roneli @bbernays).

Not sure which we should choose between TX and `ON CONFLICT UPDATE`. One thing idea in favor of using `deleteResourceByCQId` is just consistency with current behaviour in `copyFrom`. i.e., if we delete by cq-id in `copyFrom`, we should do the same kind of deletions in the fallback.

bbernays: At this point I don't think I have ever seen an issue that stems from DB performance on insert. The only incident that I am aware of is that someone had issues because they needed to vacuum their db. So overall, I am not very concerned about DB performance as our biggest bottleneck is by far and away the rate we can grab data from provider APIs...

bbernays: That being said, I like the idea and consistency of doing the deletion + insertion in a TX for each of our 3 methods of insertion (copy, insert multiple and singular insert)

database/postgres/pgdatabase.go
```diff
@@ -221,3 +228,20 @@ func quoteColumns(columns []string) []string {
 	}
 	return ret
 }
+
+type execer interface {
```
shimonp21, line 232: I'd avoid creating an "execer" type here if not absolutely necessary.
Our code base has too many "execers" already IMO.

If I understand correctly, this is only used in deleteResourceByCQId, that receives a "tx". can it just receive a plain "tx"?

disq: Fixed.

roneli (approved):
LGTM

PR #268, opened 2022-05-22
feat(resolver): Allow CIDR fallback value

erezrokah:

#### Summary

Related to https://github.com/cloudquery/cq-provider-gcp/issues/255.
The GCP API can return a valid CIDR or an empty string. This change to the SDK allows providers to have a fallback value if the parsing of the CIDR fails.

>I can add this to a bunch of other places too, but didn't do it since didn't find an actual use case for it

---

- [x] Lint your changes via `golangci-lint run --new-from-rev main` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Update or add tests 🧪

### DIFF
--- a/provider/schema/resolvers.go
+++ b/provider/schema/resolvers.go
@@ -12,6 +12,13 @@ import (
 	"github.com/thoas/go-funk"
 )
 
+type Fallback struct {
+	Value interface{}
+}
+type Options struct {
+	Fallback *Fallback
+}
+
 // PathResolver resolves a field in the Resource.Item
 //
 // Examples:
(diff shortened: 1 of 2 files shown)


provider/schema/resolvers.go
```diff
@@ -12,6 +12,13 @@ import (
 	"github.com/thoas/go-funk"
 )
 
+type Fallback struct {
+	Value interface{}
```
erezrokah, line 16: We use a separate type so we can set a `nil` fallback value

roneli:
Why not use columns Default value, the error will still exist. Moreover, I think this is a GCP specific thing, not sure if I like adding this complexity to resolvers

erezrokah:
Maybe the correct way to do it is to add it to `IgnoreError`?

erezrokah:
Ah, I think we already have a way to handle this case. Let me close and see if I can handle in `gcp` provider

PR #299, opened 2022-05-31
feat: Update default handling

roneli:

#### Summary

Update execution to set default based on resolver if it returns error or doesn't set value

- [ ] Update or add tests 🧪

### DIFF
--- a/provider/execution/execution.go
+++ b/provider/execution/execution.go
@@ -15,7 +15,6 @@ import (
 	"github.com/cloudquery/cq-provider-sdk/stats"
 	"github.com/hashicorp/go-hclog"
 	"github.com/iancoleman/strcase"
-	"github.com/modern-go/reflect2"
 	segmentStats "github.com/segmentio/stats/v4"
 	"github.com/thoas/go-funk"
 	"golang.org/x/sync/semaphore"
@@ -57,6 +56,14 @@ func NewTableExecutor(resourceName string, db Storage, logger hclog.Logger, tabl
 	if classifier != nil {
 		classifiers = append([]ErrorClassifier{classifier}, classifiers...)
 	}
+
+	// validate no pk columns have a default value defined
+	for _, pk := range table.Options.PrimaryKeys {
+		if table.Column(pk).Default != nil {
+			panic(fmt.Sprintf("pk column %s has a default value defined", pk))
+		}
+	}
+
 	var c [2]schema.ColumnList
 	c[0], c[1] = db.Dialect().Columns(table).Sift()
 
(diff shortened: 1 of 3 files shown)


shimonp21 (changes requested):
Will review again after changes

provider/execution/execution.go
```diff
@@ -407,6 +407,14 @@ func (e TableExecutor) resolveColumns(ctx context.Context, meta schema.ClientMet
 		if c.Resolver != nil {
 			e.Logger.Trace("using custom column resolver", "column", c.Name)
 			err := c.Resolver(ctx, meta, resource, c)
+			// if default wasn't set as nil and our resource is either nil or we got an error we set the default value
+			if !reflect2.IsNil(c.Default) && (err != nil || resource.Get(c.Name) != nil) {
```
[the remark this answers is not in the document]
roneli: fixed

provider/execution/execution.go
```diff
@@ -407,6 +407,14 @@ func (e TableExecutor) resolveColumns(ctx context.Context, meta schema.ClientMet
 		if c.Resolver != nil {
```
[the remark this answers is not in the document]
roneli: done

provider/execution/execution.go
```diff
@@ -407,6 +414,14 @@ func (e TableExecutor) resolveColumns(ctx context.Context, meta schema.ClientMet
...
 			e.Logger.Trace("using custom column resolver", "column", c.Name)
 			err := c.Resolver(ctx, meta, resource, c)
+			// if default wasn't set as nil and our resource is either nil or we got an error we set the default value
+			if c.Default != nil && (err != nil || resource.Get(c.Name) == nil) {
```
shimonp21, line 418: Doesn't work right now for creating an empty json for `aws_lambda_functions.Tags`.

In the golang struct from the `aws-sdk-for-go-v2`, There is a `nil` value in the `tags` field (the field is of type `map[string]string`. But in the code we converted to `interface{}`, so it becomes a `nil` of type `map[string]string`. This means that in the code when we check if `resource.Get(c.Name)` == nil, we get `false` because it's not `nil`, it's "nil of type map[string]string". Thus we don't set a default value and instead get a json-null, instead of a json-empty-objects.

Some sceenshots:

image

image

image

image

shimonp21: Rather than making the code even more complicated, e.g. with `reflect2.IsNil`, I propose to rethink whether we can just not implement this feature.

Took me 3 hours of testing to find this bug, and I am not convinced I tried all the possible configurations of nulls, types, and null-types... 

postesql nulls, nulls of type `interface{}`, nulls of other types contained in a `interface{}` (which is not null itself..)..

shimonp21: Maybe just an ad-hoc resolver for this specific function? 
The relevant issue https://github.com/cloudquery/cloudquery-issues/issues/358 just mentions 2 specific places where we don't handle nulls correctly..

PR #301, opened 2022-05-31
feat: Return full rlimit

roneli:

#### Summary

Related to https://github.com/cloudquery/cloudquery-issues/issues/372 returns full access to rlimit info on process

- [x] Update or add tests 🧪

### DIFF
--- a/helpers/limit/limits.go
+++ b/helpers/limit/limits.go
@@ -9,16 +9,21 @@ const (
 	goroutinesPerGB float64 = 250000
 )
 
+type Rlimit struct {
+	Cur uint64
+	Max uint64
+}
+
 func GetMaxGoRoutines() uint64 {
 	limit := calculateGoRoutines(getMemory())
-	ulimit, err := getUlimit()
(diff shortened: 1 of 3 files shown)


helpers/limit/limits.go
```diff
@@ -9,16 +9,21 @@ const (
...
+	ulimit, err := GetUlimit()
+	if err != nil || ulimit.Max == 0 {
 		return limit
 	}
-	if ulimit > limit {
+	if ulimit.Max > limit {
 		return limit
 	}
-	return ulimit
+	return ulimit.Max
```
bbernays, line 26: On a seperate/ related note, instead of going all the way up to the `ulimit` why not do something like `ulimit * 0.8` So then we don't try and take all of the descriptors because we know there are other things that use file descriptors other than our go routines

erezrokah (approved):

helpers/limit/limits.go
roneli: We can do that, will do in a separate pr

PR #321, opened 2022-06-07
feat: Always use BigInt

yevgenypats:

#### Summary

This is a pre-step to supporting only `TypeBigInt`. we don't need necessarly to remove this as this will break a bit of backward compatibility but at least right now we shouldn't have any int mismatch errors as we always use BigInt.

---

- [ ] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [ ] Update or add tests 🧪

### DIFF
--- a/provider/schema/column.go
+++ b/provider/schema/column.go
@@ -98,12 +98,8 @@ func (v ValueType) String() string {
 	switch v {
 	case TypeBool:
 		return "TypeBool"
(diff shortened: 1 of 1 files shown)


yevgenypats:
Also, found a small bug where we didn't support `*uint16`

disq (approved):
LGTM

PR #283, opened 2022-05-24
fix(testing): Don't add ignored diagnostics to errors validation

erezrokah:
🎉 Thank you for making CloudQuery awesome by submitting a PR 🎉

#### Summary

Related to https://github.com/cloudquery/cq-provider-gcp/pull/267.
The tests on https://github.com/cloudquery/cq-provider-gcp/pull/267 are failing since we validate ignored errors.
Relevant SDK code is https://github.com/cloudquery/cq-provider-sdk/blob/849cad30f6c3b335981822f19fdf275f6091df03/provider/execution/execution.go#L412

---

Use the following steps to ensure your PR is ready to be reviewed

- [x] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [x] Run `go fmt` to format your code 🖊
- [x] Lint your changes via `golangci-lint run --new-from-rev main` 🚨 (install golangci-lint [here](https://golangci-lint.run/usage/install/#local-installation))
- [x] Update or add tests 🧪
- [x] Ensure the status checks below are successful ✅

PR #323, opened 2022-06-07
feat!: IgnoreError Recursively for tables and columns

yevgenypats:

#### Summary

This adds recursive logic for both columns and tables so we dont have to put it in every column and relation. 

This need more tests but ready for initial review.

Closes https://github.com/cloudquery/cq-provider-sdk/pull/323

---

- [ ] Read the [contribution guidelines](../blob/main/.github/CONTRIBUTING.md) 🧑‍🎓
- [ ] Update or add tests 🧪

### DIFF
--- a/provider/execution/execution.go
+++ b/provider/execution/execution.go
@@ -414,7 +414,7 @@ func (e TableExecutor) resolveColumns(ctx context.Context, meta schema.ClientMet
 				return diags.Add(ClassifyError(err, diag.WithResourceName(e.ResourceName), WithResource(resource), diag.WithSummary("failed to resolve column %s@%s", e.Table.Name, c.Name)))
 			}
 			// check if column resolver defined an IgnoreError function, if it does check if ignore should be ignored.
-			if e.Table.IsIgnoreErrorColumn(c, err) {
+			if e.Table.IsIgnoreError(err) {
 				diags = diags.Add(e.handleResolveError(meta, resource, err, diag.WithSeverity(diag.IGNORE), diag.WithSummary("column resolver %q failed for table %q", c.Name, e.Table.Name)))
 			} else {
 				diags = diags.Add(e.handleResolveError(meta, resource, err, diag.WithSummary("column resolver %q failed for table %q", c.Name, e.Table.Name)))
--- a/provider/execution/execution.go
+++ b/provider/execution/execution.go
@@ -92,6 +92,7 @@ func (e TableExecutor) withTable(t *schema.Table, kv ...interface{}) *TableExecu
 	var c [2]schema.ColumnList
 	c[0], c[1] = e.Db.Dialect().Columns(t).Sift()
 	cpy := e
+	cpy.Table.Parent = t.Parent
 	cpy.Table = t
 	cpy.Logger = cpy.Logger.With(kv...)
 	cpy.columns = c
@@ -236,7 +237,7 @@ func (e TableExecutor) callTableResolve(ctx context.Context, client schema.Clien
 			close(res)
 		}()
 		if err := e.Table.Resolver(ctx, client, parent, res); err != nil {
-			if e.Table.IgnoreError != nil && e.Table.IgnoreError(err) {
+			if e.Table.IsIgnoreError(err) {
 				e.Logger.Debug("ignored an error", "err", err)
 				err = diag.NewBaseError(err, diag.RESOLVING, diag.WithSeverity(diag.IGNORE), diag.WithSummary("table %q resolver ignored error", e.Table.Name))
 			}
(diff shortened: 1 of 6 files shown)


provider/schema/table.go
```diff
@@ -80,6 +82,40 @@ func (tco TableCreationOptions) signature() string {
...
+	if t.Parent != nil {
+		return t.Parent.IsIgnoreError(e)
+	}
+
+	return false
+}
+
+// IsIgnoreErrorColumn returns true if the error is ignored via the column resolver, current table IgnoreError function or in any other parent table (in that ordered)
+// it stops checking the moment one of them exists and not until it returns true or fals
+func (t Table) IsIgnoreErrorColumn(c Column, e error) bool {
```
erezrokah, line 102: What is the use case for running the table ignore error function for specific columns?
Are we expecting similar errors to happen?

When I added an ignore error for columns it was super specific to the data type of the column and the provider API:
https://github.com/cloudquery/cq-provider-gcp/blob/49238085984bc25468ced7484761a3ceb1b7a99e/client/error.go#L59

I'd expect that table level errors are more about failures to fetch the data, and column level errors are more about failures to parse/normalize the data

erezrokah (commented):
I'll let others reviews comment too, but WDYT about starting with the table level changes, and then moving to the column level changes?

provider/schema/table.go
yevgenypats: Where do we use ColumnIgnoreError apart from this case? trying to find it with regex...

erezrokah: I think I might have been the first consumer of that feature as I had to do another fix in our testing logic to get it to work https://github.com/cloudquery/cq-provider-sdk/pull/283

yevgenypats: I see. so seems to me it will be best to remove this feature as well as it is unused, so less documentation, less bugs. and given this PR was the first consumer I suggest to implement it by wrapping up `IPNetResolver` with `GCPIPNetResolver` that will also ignore this error. WDYT ?

bbernays: So why not just only recursively propagate the table level error handler?

erezrokah: Ah well, that's how I started and the realized we have in the SDK. I'll open a new issue for that.

yevgenypats: Or even better add to the `IPNetResolver` a check that if it's empty just don't do anything. seems to me like a good default behaviour for all cases or just set it to null instead. both seems good to me.

erezrokah: Opened https://github.com/cloudquery/cq-provider-sdk/issues/327 and https://github.com/cloudquery/cq-provider-sdk/issues/326.

Didn't want to change the default `IPNetResolver` back then as I only had a very specific use case for the GCP provider. Didn't see empty strings coming from other APIs.

yevgenypats: I actually already pushed it as it's less code in this PR, so I just need to add a test now

provider/execution/execution.go
```diff
@@ -92,6 +92,7 @@ func (e TableExecutor) withTable(t *schema.Table, kv ...interface{}) *TableExecu
 	var c [2]schema.ColumnList
 	c[0], c[1] = e.Db.Dialect().Columns(t).Sift()
 	cpy := e
+	cpy.Table.Parent = t.Parent
```
disq, line 95: You don't need this since you have line 96?

erezrokah (commented):
Add two more comments

provider/execution/execution.go
yevgenypats: I think I do because parent is not set in table automatically (i.e by the user). I can also move this function to execution and not to table if that will make more sense.

provider/execution/execution.go
```diff
@@ -413,7 +414,7 @@ func (e TableExecutor) resolveColumns(ctx context.Context, meta schema.ClientMet
 				return diags.Add(ClassifyError(err, diag.WithResourceName(e.ResourceName), WithResource(resource), diag.WithSummary("failed to resolve column %s@%s", e.Table.Name, c.Name)))
 			}
 			// check if column resolver defined an IgnoreError function, if it does check if ignore should be ignored.
-			if c.IgnoreError != nil && c.IgnoreError(err) {
+			if e.Table.IsIgnoreError(err) {
```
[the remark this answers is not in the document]
yevgenypats: Good point. Yeah we can do that. It might be needed in the future but Im all in on removing now and adding in the future if needed.

provider/execution/execution.go
```diff
@@ -236,7 +237,7 @@ func (e TableExecutor) callTableResolve(ctx context.Context, client schema.Clien
 			close(res)
 		}()
 		if err := e.Table.Resolver(ctx, client, parent, res); err != nil {
-			if e.Table.IgnoreError != nil && e.Table.IgnoreError(err) {
+			if e.Table.IsIgnoreError(err) {
```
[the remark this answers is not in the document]
yevgenypats: I think there is already a debug log there on line 241

erezrokah (approved):
🚀  @yevgenypats please squash merge with a commit title prefix of `feat!: ` as this is a breaking change

yevgenypats:
@erezrokah will do! Will wait on @disq approval (regarding comment) and will squash merge 🚀

provider/execution/execution.go
disq: But this is overridden by L96, so it has no effect?

disq (approved):
LGTM

PR #332, opened 2022-06-13
feat!: Support both YAML and HCL config

disq:


provider/provider.go
```diff
@@ -113,31 +140,36 @@ func (p *Provider) ConfigureProvider(_ context.Context, request *cqproto.Configu
...
+			}
+		}
+
+		client, diags := p.Configure(p.Logger, providerConfig)
+		if diags.HasErrors() {
+			return &cqproto.ConfigureProviderResponse{
+				Diagnostics: diags,
+			}, nil
+		}
 	}
```
disq, line 173: TODO do the same for ConfigYAML

provider/provider.go
```diff
@@ -113,23 +140,32 @@ func (p *Provider) ConfigureProvider(_ context.Context, request *cqproto.Configu
...
+
 	providerConfig := p.Config()
 	if err := defaults.Set(providerConfig); err != nil {
 		return &cqproto.ConfigureProviderResponse{
 			Diagnostics: diag.FromError(err, diag.INTERNAL),
 		}, nil
 	}
+
 	// if we received an empty config we notify in log and only use defaults.
 	if len(request.Config) == 0 {
```
roneli, line 152: Looking good, some thoughts, tell me what you think about them:

why not let in the request the core passes the format of configuration? prior cores by default its set to HCL and used. new cores can decide based on configuration passed. 

Should the "provider" care / pass what format is supports? it just needs to know how to decode the block and pass as interface to main Configure function. I think also in the GetProviderConfig request we should do the same we can request yaml/hcl format if provider doesn't support yaml format we default back to hcl.

roneli (commented):
Looking good, didn't approve as this is a draft, added some of my thoughts, tell me what you think :)

roneli (commented):
Looks good, obviously not finished, but didn't have much comments one nit. yaml.Node is a great  solution.

provider/provider.go
```diff
@@ -71,19 +84,48 @@ func (p *Provider) GetProviderSchema(_ context.Context, _ *cqproto.GetProviderSc
...
 	}
-	data := fmt.Sprintf(`
+	switch providerConfig.Format() {
+	case cqproto.ConfigHCL:
+		data := fmt.Sprintf(`
 		provider "%s" {
 			%s
 			// list of resources to fetch
 			resources = %s
-		}`, p.Name, p.Config().Example(), helpers.FormatSlice(funk.Keys(p.ResourceMap).([]string)))
```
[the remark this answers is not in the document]
disq: IMO that would affect readability adversely (if const it has to be up top, we have multiple params, etc)

provider/provider.go
```diff
@@ -71,19 +78,70 @@ func (p *Provider) GetProviderSchema(_ context.Context, _ *cqproto.GetProviderSc
...
+		}`, p.Name, providerConfig.Example(), helpers.FormatSlice(funk.Keys(p.ResourceMap).([]string)))
+
+		return &cqproto.GetProviderConfigResponse{
+			Config: hclwrite.Format([]byte(data)),
+			Format: cqproto.ConfigHCL,
+		}, nil
+	case cqproto.ConfigYAML:
+		var data ProviderConfiguration
+		if err := yaml.Unmarshal([]byte(providerConfig.Example()), &data.Inline); err != nil {
```
yevgenypats, line 101: why the unmarshal/marshal is necessary here? why not just send the []byte/string as is?

yevgenypats (approved):
Looks good. just had one question/clarification.

provider/provider.go
disq: because we add resource list to the bottom in line 105..130

provider/provider.go
```diff
@@ -71,19 +78,59 @@ func (p *Provider) GetProviderSchema(_ context.Context, _ *cqproto.GetProviderSc
...
-		}`, p.Name, p.Config().Example(), helpers.FormatSlice(funk.Keys(p.ResourceMap).([]string)))
+		}`, p.Name, providerConfig.Example(), helpers.FormatSlice(funk.Keys(p.ResourceMap).([]string)))
+
+		return &cqproto.GetProviderConfigResponse{
+			Config: hclwrite.Format([]byte(data)),
+			Format: cqproto.ConfigHCL,
+		}, nil
+	case cqproto.ConfigYAML:
+		var data ProviderConfiguration
+		if err := yaml.Unmarshal([]byte(providerConfig.Example()), &data); err != nil {
```
yevgenypats, line 101: Why this unmarshal/marshal is necessary ?

roneli (approved):
LGTM, no further comments :)

PR #363, opened 2022-06-20
fix: Use errgroup SetLimit

erezrokah:

#### Summary

With [sync package update](https://github.com/cloudquery/cq-provider-sdk/pull/345) `errgroup` has support for setting a limit.
We should be able to do the same in other repositories.
Still need to test it a bit.

---

- [x] Update or add tests 🧪

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -186,14 +186,10 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
 	p.Logger.Info("calculated max goroutines for fetch execution", "max_goroutines", maxGoroutines)
 	goroutinesSem = semaphore.NewWeighted(helpers.Uint64ToInt64(maxGoroutines))
 
-	// limiter used to limit the amount of resources fetched concurrently
-	var parallelResourceSem *semaphore.Weighted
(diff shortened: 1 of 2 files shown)


provider/provider.go
```diff
@@ -186,14 +186,10 @@ func (p *Provider) FetchResources(ctx context.Context, request *cqproto.FetchRes
...
 
-	// limiter used to limit the amount of resources fetched concurrently
-	var parallelResourceSem *semaphore.Weighted
-	maxParallelFetchingLimit := request.ParallelFetchingLimit
-	if maxParallelFetchingLimit > 0 {
-		parallelResourceSem = semaphore.NewWeighted(helpers.Uint64ToInt64(maxParallelFetchingLimit))
-	}
-
 	g, gctx := errgroup.WithContext(ctx)
+	if request.ParallelFetchingLimit > 0 {
```
erezrokah, line 190: Need to check if we can drop this condition by verifying that `0 == no limit`

erezrokah:
Tested it and it seems to work as expected

shimonp21:
Awesome refactor - thanks!

shimonp21 (approved):

PR #371, opened 2022-06-24
fix: Issues when PG username is 'cloudquery'

### LINKED ISSUE #901: Specify Schema
### Describe the Bug

Right now it's not possible to specify schema when we define the database. It mostly uses the default `search_path` which in some situations can be not in the user control depending on which database it uses. 

Issue noted in GCP on Cloud SQL creating all tables in `cloudquery` schema instead of `public`

shimonp21:
Before this fix, cloudquery misbehaved when the PG username was named 'cloudquery'.
It created tables in the 'cloudquery' schema instead of the 'public' schema.
This happened because the search path was  '"$user", public', which means it depended on the username.

This fix, upon connection to PG, will change serach_path to 'public' only, i.e. not dependent on username (which is way more predictable).

This won't impact us explicitly accessing the 'cloudquery' schema via 'cloudquery.<table-name>', which is what we do when we actually need the 'cloudquery' schema.

closes: https://github.com/cloudquery/cloudquery/issues/901.

#### Summary

---

- [x] Update or add tests 🧪

### DIFF
--- a/database/postgres/connection.go
+++ b/database/postgres/connection.go
@@ -16,6 +16,11 @@ func Connect(ctx context.Context, dsnURI string) (*pgxpool.Pool, error) {
 		return nil, dsn.RedactParseError(err)
 	}
 	poolCfg.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
+		_, err := conn.Exec(ctx, "SET search_path=public")
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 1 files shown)


yevgenypats (approved):
Just make sure it wont put the cloudquery schema in public.

shimonp21:
Made sure :)

Issue #273, opened 2022-05-23
Cleanup global tables


shimonp21:
after #266 global tables are redundant and should be deleted.

disq:
Maybe not, depending on what we (possibly) do with the temp table cq_id dupe detection.

shimonp21:
We should also consider deleting "deleteCascadeFilters"

disq:
Also `ExtraFields` and `AlwaysDelete` (as mentioned in https://github.com/cloudquery/cq-provider-sdk/pull/265 )

PR #392, opened 2022-06-28
chore!: Remove unused code/features: Global tables, CascadeDeleteFilters, ExtraFields, AlwaysDelete

disq:
Completes https://github.com/cloudquery/cq-provider-sdk/issues/308

#273 #309 #310 #311


yevgenypats (approved):
Love code deletion!

cqproto/internal/plugin.proto
```diff
@@ -24,8 +24,6 @@ message ConfigureProvider {
     ConnectionDetails connection = 2;
     // Holds information such as credentials, regions, accounts, etc'
     bytes config = 3;
-    // Msgpack encoded: Allows to inject & override fields into resource tables, use this carefully to override fields. Used in tests.
-    bytes extraFields = 5;
```
roneli, line 28: not 100% sure, but wouldn't this break prior versions? wouldn't it be better to do [deprecated = true] instead, and when decoding / encoding with new SDK ignore it?

provider/schema/table.go
```diff
@@ -43,19 +43,13 @@ type Table struct {
...
 
 	// IgnoreInTests is used to exclude a table from integration tests.
 	// By default, integration tests fetch all resources from cloudquery's test account, and verifY all tables
 	// have at least one row.
 	// When IgnoreInTests is true, integration tests won't fetch from this table.
 	// Used when it is hard to create a reproducible environment with a row in this table.
 	IgnoreInTests bool
-	// Global tables are usually the same regardless of the provider fetch configuration. Global table data gets fetched
-	// and doesn't produce PK conflict errors instead data is replaced
-	Global bool
```
roneli, line 58: nit: would require a PR to AWS provider, same with AlwaysDelete

roneli (approved):
LGTM, had two comments. The second is stating the obvious ^^, so just auto-resolve after reading it.

provider/schema/table.go
disq: https://github.com/cloudquery/cq-provider-aws/pull/1124

cqproto/internal/plugin.proto
disq: it would just show up as default empty value which would be a nil slice and map to a nil map.

PR #400, opened 2022-07-01
feat(tests): Fetch only the resources required for test being run

### LINKED ISSUE #398: feat(integration tests): Integration tests should only fetch the relevant resources
When running the integration test suite for a specific table, e.g. 

```
go test -run="TestIntegration/aws_lightsail_instances" -tags=integration ./...
```

it should fetch only the relevant resources. Currently, the test will fetch all resources but then only run assertions against the given table(s). This can take a few minutes, and it would be much faster if we can limit resource fetching to only the specified service(s).

hermanschaaf:
#### Summary

This closes https://github.com/cloudquery/cq-provider-sdk/issues/398

When adding a new resource, it is desirable to only fetch that resource during integration testing. Before this change, all resources were fetched (which took about 3.5 minutes). After the change, only the resources required by the subtest are fetched (so test runs for cq-provider-aws now take about 10 seconds on my machine, when limiting to a specific table). 

For example, when running integration tests with:

```
go test -v -run="TestIntegration/aws_lightsail_instances" -tags=integration ./...
```

only the resources required to populate the `aws_lightsail_instances` and related tables will be fetched. If a subtest is not being run, all resources will still be fetched. This means integration tests for a specific resource can be run in about 10 seconds, as opposed to the several minutes it took before. The trade-off is, however, that the full suite will take slightly longer to run. I think this is worthwhile trade-off to make, as long as we're not also jeopardizing correctness. I think this could happen if there are dependencies between resources, but hopefully the tests for this PR will fail now if this is the case.

---

- [x] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [x] Update or add tests 🧪

### DIFF
--- a/provider/schema/table.go
+++ b/provider/schema/table.go
@@ -48,7 +48,7 @@ type Table struct {
 	AlwaysDelete bool
 
 	// IgnoreInTests is used to exclude a table from integration tests.
-	// By default, integration tests fetch all resources from cloudquery's test account, and verifY all tables
+	// By default, integration tests fetch all resources from cloudquery's test account, and verify all tables
 	// have at least one row.
 	// When IgnoreInTests is true, integration tests won't fetch from this table.
 	// Used when it is hard to create a reproducible environment with a row in this table.
--- a/provider/testing/resource.go
+++ b/provider/testing/resource.go
@@ -88,7 +88,7 @@ func TestResource(t *testing.T, resource ResourceTestCase) {
 	}
 
 	for resourceName, table := range resource.Provider.ResourceMap {
-		t.Run(table.Name, func(t *testing.T) {
+		t.Run(resourceName, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


yevgenypats:
Looks good! Maybe we should pass the resource name instead the table name to `t.Run` - this should make things simpler and will avoid any reverse-lookup for table <-> resource because there can be multiple nested tables in one resource

erezrokah (approved):
Amazing 🔥 To fix the lint errors you'll need to run `gci write .` (https://github.com/daixiang0/gci)

hermanschaaf:
> Looks good! Maybe we should pass the resource name instead the table name to t.Run 

@yevgenypats I've made this change now. We'll have to update some examples in some plugin repos to use the new format, I'll open new PRs for those.

PR #415, opened 2022-07-07
fix: Optional forced search_path

disq:
Should go with https://github.com/cloudquery/cloudquery/pull/1112 .

Partially reverts https://github.com/cloudquery/cq-provider-sdk/pull/371 (depending)

### DIFF
--- a/database/postgres/connection.go
+++ b/database/postgres/connection.go
@@ -2,6 +2,7 @@ package postgres
 
 import (
 	"context"
+	"strings"
 
 	"github.com/cloudquery/cq-provider-sdk/database/dsn"
 	"github.com/jackc/pgtype"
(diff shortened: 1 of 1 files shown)


database/postgres/connection.go
```diff
@@ -16,9 +17,12 @@ func Connect(ctx context.Context, dsnURI string) (*pgxpool.Pool, error) {
 		return nil, dsn.RedactParseError(err)
 	}
 	poolCfg.AfterConnect = func(ctx context.Context, conn *pgx.Conn) error {
-		_, err := conn.Exec(ctx, "SET search_path=public")
-		if err != nil {
-			return err
+		// force a known search_path if DSN doesn't specify one
+		if !strings.Contains(dsnURI, "&search_path=") && !strings.Contains(dsnURI, "?search_path=") {
```
irmatov, line 21: maybe it's better to parse dsn as URI instead of trying a simple text search?

hermanschaaf, line 21: Doesn't matter much, but how would you feel about a regex here? Something like:

```
var containsSearchPath  = regexp.MustCompile(`[?&]search_path\=`).MatchString
...
if !containsSearchPath(dnsURI) {
  ...
}
```

hermanschaaf (approved):

database/postgres/connection.go
disq: Would add complexity and we only want to check if they have a `search_path` in there.

PR #418, opened 2022-07-11
fix!: Diagnostic cleanup

shimonp21:
#### Summary
[This PR is split into multiple commits, please review them separately]

This commit includes some significant improvements to how error-handling is done in cloudquery.
Most importantly - it disallows conversion of Diagnostics (or 'Diagnostic', or 'BaseError') to 'error'.
Implicit conversion between diags and errors caused many hard to track bugs (e.g. an empty list of diagnostics becoming
an 'error: no errors').

This commit also:
- Removes 'IgnoreError' completely. This was kind of dead code - errors marked as ignored could be overwritten and become warnings if the ErrorClassidfier classified them as warnings. This happened through a very convoluted flow converting between diags and errors several times, causing messages such as "warning: cloudquery ignored an error: access denied", instead of a much more straightforward "warning: access denied".
  - When upgrading a provider to this SDK version, make sure to move any 'ignores' that you require to the ErrorClassifier (i.e. classify them as 'IGNORED' severity).

- Changes 'ErrorClassifier' to just receive a single 'error'. It moves all the implicit information that uesd to be passed by creating a diag with context (e.g. resourceID) and passing it to the classifier, to just being explicit parameters. Some more cleanup may be a good idea here. So, the task of an errorClassifier is more straigtforward now - receive an error, convert it to a diagnostic.

See the corresponding cq-provider-aws PR for an example of how upgrade the SDK to this version.

---

- [ ] Update or add tests 🧪


provider/diag/diagnostics.go
```diff
@@ -15,7 +15,10 @@ type Diagnostics []Diagnostic
 
 var _ sort.Interface = (*Diagnostics)(nil)
 
-func (diags Diagnostics) Error() string {
+// Deprecated! Only here for backward compatibility (specifically in the gRPC protocol). Do not use unless
+//  you are 100% sure you need this.
+// Converts a list of diagnostics into an error.
+func (diags Diagnostics) DeprecatedToError() string {
```
disq, line 21:
```suggestion
func (diags Diagnostics) ToError() string {
```
Would just mark it as `Deprecated:` as described in [here](https://stackoverflow.com/questions/7849663/how-do-you-mark-code-as-deprecated-in-go) (The `Deprecated!` usage above is also non standard)

yevgenypats:
I'll wait a bit with this PR to give a shot at removing Diags completely and only using logging mechanism.

yevgenypats:
Closing in favour of https://github.com/cloudquery/cq-provider-sdk/pull/437

PR #419, opened 2022-07-13
fix: Remove dead code

shimonp21:
#### Summary

The code inside the defaultErrorClassifier was never actually executed, because it was always passed a 'diag', but the first line in defaultErrorClassifier was "if err type is diag, return nil".

In any case, it's definitely better not to classify a single error twice. A more reasonable implementation might be to 'fall back to default classifier, if specific classifier returned nil', but for now, let's just remove it.

---

- [ ] Update or add tests 🧪


roneli (approved):
LGTM

candiduslynx (approved):

PR #424, opened 2022-07-19
chore!: Remove HCL config support

disq:
Remove some dead code we don't intend to develop any more.


cqproto/internal/plugin.proto
```diff
@@ -134,20 +133,15 @@ message GetProviderSchema {
...
 message GetProviderConfig {
   message Request {
-    ConfigFormat format = 1;
+    // next available index: 2
   }
   message Response {
     string name = 1;
     string version = 2;
     bytes config = 3;
-    ConfigFormat format = 4;
```
roneli, line 150: maybe keep this but mark it as deprecated

roneli (approved):
LGTM, had one nit

cqproto/internal/plugin.proto
disq: In the response only?

yevgenypats (approved):

PR #425, opened 2022-07-19
feat: Better backwards compatibility with older providers/core

disq:
Always include config format in requests to provider. (apparently removed too much in https://github.com/cloudquery/cq-provider-sdk/pull/424 )


cqproto/internal/plugin.proto
```diff
@@ -135,19 +136,21 @@ message GetProviderSchema {
...
+    // Deprecated. Format of the config, always ConfigYAML
+    ConfigFormat format = 1;
   }
   message Response {
     string name = 1;
     string version = 2;
     bytes config = 3;
-    ConfigFormat format = 4; // deprecated
+    // Deprecated. Format of the config, always ConfigYAML
+    ConfigFormat format = 4;
```
roneli, line 153: You have a deprecated keyword in proto i thibk maybe set it

roneli (approved):
Lgtm 1 nit

PR #435, opened 2022-07-22
feat!: Validate provider config

disq:
Implements https://github.com/cloudquery/cloudquery/issues/1104
I'm not sure if this is really necessary at this point, as we decode the config before validating schema anyway. If it doesn't decode cleanly we just stop there and don't go into the schema validation.

This PR also starts decoding the config in strict mode, unknown keys in the config will produce errors from now on.


provider/provider.go
```diff
@@ -145,28 +207,48 @@ func (p *Provider) ConfigureProvider(_ context.Context, request *cqproto.Configu
 		}
 	}
 
+	var diags diag.Diagnostics
+
+	// validate provider configuration before proceeding
+	validateResponse, err := p.ValidateProviderConfig(ctx, &cqproto.ValidateProviderConfigRequest{Config: request.Config})
```
disq, line 213: Every call to `ConfigureProvider` also runs `ValidateProviderConfig`. This also means we decode and defaultize the config twice: once in validate and one down below.

yevgenypats:
Maybe worth doing a live walkthrough on this to go through all the cases. One  questions in the meanwhile:
1) why not just send the schema json to core from the the provider and make the core do the validation? this will make remove the need to send all the diags stuff from provider to core and we can use normal errors?

yevgenypats:
Im working on a complete re-write of cloudquery core including no-diags so let's put this on hold for now as otherwise I wont be able to merge:
https://github.com/cloudquery/cloudquery/pull/1139
https://github.com/cloudquery/cq-provider-sdk/pull/437

yevgenypats (changes requested):
Going into a feature freeze in cloudquery core and sdk apart from P0 bugs until cloudquery v2 is done https://github.com/cloudquery/cloudquery

PR #436, opened 2022-07-23
fix: Fixed print of diag errors

amanenk:
#### Summary

fixed print of provider configuration error when error is in diags.

---

- [x] Read the [contribution guidelines](../blob/main/CONTRIBUTING.md) 🧑‍🎓
- [x] Update or add tests 🧪

### DIFF
--- a/provider/testing/resource.go
+++ b/provider/testing/resource.go
@@ -84,7 +84,7 @@ func TestResource(t *testing.T, resource ResourceTestCase) {
 	}); configErr != nil {
 		t.Fatal("failed to configure provider", configErr)
 	} else if resp != nil && resp.Diagnostics.HasErrors() {
-		t.Fatal("errors while configuring provider", configErr)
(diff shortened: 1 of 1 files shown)


provider/testing/resource.go
```diff
@@ -84,7 +84,7 @@ func TestResource(t *testing.T, resource ResourceTestCase) {
 	}); configErr != nil {
 		t.Fatal("failed to configure provider", configErr)
 	} else if resp != nil && resp.Diagnostics.HasErrors() {
-		t.Fatal("errors while configuring provider", configErr)
+		t.Fatal("errors while configuring provider", resp.Diagnostics.Error())
```
amanenk, line 87: should we print only errors in `resp.Diagnostics.Error` function or it is ok to print all the `diags`?

yevgenypats:
We are going to remove diags. so let's hold with any code that touches it atm. there are a lot of bugs related to that but they will partially go away with https://github.com/cloudquery/cq-provider-sdk/pull/437

PR #437, opened 2022-07-23
Refactor/no diags

yevgenypats:
This is very much wip but this is removing use of diags which due to the nature of it "infects" the whole code-bases similar to `context`. We are moving to pure logging as it is already structured and it is being send anyway to the core by the hashicorp go-plugin library.

---

- [ ] Update or add tests 🧪


provider/execution/execution.go
```diff
@@ -178,45 +165,37 @@ func (e TableExecutor) cleanupStaleData(ctx context.Context, client schema.Clien
...
 	}
 	e.Logger.Debug("cleaned table stale data successfully", "last_update", e.executionStart)
-	return nil
+	return
 }
 
 // callTableResolve does the actual resolving of the table calling the root table's resolver and for each returned resource resolves its columns and relations.
-func (e TableExecutor) callTableResolve(ctx context.Context, client schema.ClientMeta, parent *schema.Resource) (uint64, diag.Diagnostics) {
-	clock := stats.NewClockWithObserve("callTableResolve", segmentStats.Tag{Name: "client_id", Value: identifyClient(client)}, segmentStats.Tag{Name: "table", Value: e.Table.Name})
-	defer clock.Stop()
```
roneli, line 191: why did u remove the stats? this is so we can see when we have hanging resolvers that are still running

roneli (commented):
Had one comment, regarding the proto, make sure running a new provider version will work with a diag based core, i.e works but doesn't print diags. Unless we don't want that so in that case increase the proto version so there will be a mismatch and users will have to upgrade.

PR #440, opened 2022-07-26
feat: Add new Resolver PathTableResolver

### LINKED ISSUE #432: path resovler for relations
create a function that works as path resolver for columns but for relations.

instead of filling up a resolver with repeatable code like below 
```
func fetchLightsailInstanceAddOns(ctx context.Context, meta schema.ClientMeta, parent *schema.Resource, res chan<- interface{}) error {
	instance := parent.Item.(types.Instance)
	res <- instance.AddOns
	return nil
}
```
That would be nice to just set a resovler 
```
{
				Name:          "aws_lightsail_instance_add_ons",
				Description:   "Describes an add-on that is enabled for an Amazon Lightsail resource.",
				//Resolver:      fetchLightsailInstanceAddOns,
				Resolver:      schema.RelationPathResolver("AddOns"),
				IgnoreInTests: true,
				Columns: []schema.Column{
				...
```

bbernays:
#### Summary

I think this actually implements and closes #432 

---

- [ ] Update or add tests 🧪


candiduslynx (approved):

disq (approved):

provider/schema/resolvers_test.go
```diff
@@ -130,24 +131,74 @@ var UUIDTestTable = &Table{
...
+			expectedValue: innerStruct{Value: "bla"},
+		},
+		{
+			path:          "Inner.Value",
+			expectedValue: "bla",
+		},
+		{
+			path:          "NilInner.Value",
+			expectedValue: "",
+		},
```
hermanschaaf, line 151: Should we also add a case for the unexported value? What is expected to happen in that case?

bbernays: Added a test for that

hermanschaaf (approved):

